PR #107, opened 2026-03-23
feat: complete profile flow

gulcetahtasiz:
This PR implements the complete profile flow and profile view. It introduces the /complete-profile page where users can enter their physical information, medical history, and detailed address (country, city, district, neighborhood, and extra address) with dynamic location selection. It also adds the /profile page to display and update this information.

The authentication flow is updated so that after successful email verification, users are redirected to the complete profile page. Additionally, the signup flow now preserves email and phone data so they can be merged into the profile later.

Profile data is stored in localStorage under the key “user”. During save, existing user data (such as email and phone) is merged with the newly entered profile data, and the profile page reads from the same source.

Changes are scoped to feature-level components and pages, and shared UI components were not modified to avoid breaking existing structure.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -1,4 +1,9 @@
 {
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start"
+  },
   "dependencies": {
     "clsx": "^2.1.1",
     "next": "^16.2.0",


erinc00 (approved):
Looks good to me now. The main issues from the previous review appear to be addressed, and I do not see any major remaining blocker from this diff.

PR #174, opened 2026-03-29
fix: integrate backend with android and rearrange signup and complete profile flows

### LINKED ISSUE #132: [FEATURE] Integrate frontend with backend APIs
### Primary Objective

Connect frontend components with backend endpoints

### Motivation

Ensure real data flow between frontend and backend.

### Scope

- Endpoint integration
- Error handling

### Feature Details

Connect frontend components with backend endpoints.
Handle data fetching, updates, and basic error states.

### Acceptance Criteria

- [x] Code implemented
- [ ] Unit tests passed
- [ ] Documentation updated

### Notes

- Ensure payload compatibility

kckagancan:
## Summary

This PR integrates the backend with Android clients for the authentication and profile flows.

It also updates the signup flow so that users no longer enter profile details such as full name and phone number during signup. These fields are now collected later in the complete profile flow, making the authentication flow simpler and keeping profile completion as a separate step.

## Changes

- integrated backend API calls for Android auth/profile flow
- connected login, signup, email verification, profile fetch, and profile update flows to the backend
- moved full name and phone number collection out of signup for both web and android (1 page for web)
- updated complete profile flow to collect the missing required profile fields
- updated profile/edit profile flow to sync with backend data

## Notes

- signup now focuses only on account creation
- profile completion is handled after authentication
- both web and Android now follow the same general auth/profile flow
- only one file from web/ is changed. Rest of the changes are for android/

Can you review? @erinc00 @gulcetahtasiz 

Closes #132

### DIFF
--- a/android/app/src/debug/AndroidManifest.xml
+++ b/android/app/src/debug/AndroidManifest.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <application
+        android:usesCleartextTraffic="true" />
+
+</manifest>
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -6,7 +6,6 @@
     <application
         android:allowBackup="true"
         android:label="@string/app_name"
-        android:usesCleartextTraffic="true"
         android:supportsRtl="true"
         android:theme="@style/Theme.Neph">
 
--- a/android/app/src/main/java/com/neph/features/auth/data/AuthRepository.kt
+++ b/android/app/src/main/java/com/neph/features/auth/data/AuthRepository.kt
@@ -79,8 +79,6 @@ object AuthRepository {
             AuthSessionStore.clearPendingVerificationEmail()
             LoginDestination.PROFILE
         } catch (cancellationException: CancellationException) {
-            AuthSessionStore.clearAccessToken()
-            ProfileRepository.clearProfile()
             throw cancellationException
         } catch (error: ApiException) {
             when (error.status) {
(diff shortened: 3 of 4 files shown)


kckagancan:
Thanks for the detailed review. Went through the points and updated the Android side accordingly.

* **AuthRepository.login()**
  Adjusted the login flow so that a successful auth session is no longer cleared for temporary/non-auth profile fetch failures. Session cleanup is now limited to auth-invalid cases such as **401**. We also stopped clearing the session on coroutine cancellation.

* **SignupScreen**
  Re-checked this flow and did not find a pre-success local profile write in `SignupScreen`. The signup screen only triggers `AuthRepository.signup()` and proceeds on success. The local profile write happens after a successful signup response.

* **usesCleartextTraffic**
  Removed the global `usesCleartextTraffic="true"` from the main manifest and scoped it to the **debug** manifest only, so local HTTP development still works without leaving the setting broadly enabled.

* **ProfileRepository.syncProfile()**
  The concern about fragmented multi-endpoint updates is valid. Given the current backend contract, the client still has to perform multiple sequential updates, so a fully atomic fix is not possible on Android alone. As a mitigation, we improved the failure path to re-fetch and re-align local profile state with the backend when a partial failure occurs.

* **Birth date vs age**
  Agreed that this is primarily a backend contract follow-up. The Android UI currently collects `birthDate`, but the backend persists only `age`, so a proper round-trip `birthDate` field would need backend support.

Can you review again? @erinc00

gulcetahtasiz (approved):
Looks good now, thanks for addressing the feedback.

The issues raised in the previous review are properly resolved, especially around session handling, signup flow, and cleartext configuration. The syncProfile mitigation also makes sense given current backend constraints.

The remaining birthDate note can be handled on the backend side. approved

Issue #193, opened 2026-04-02
[FEATURE] Implement Android Help Request and Availability Flows


gulcetahtasiz:
### Primary Objective

Complete the Android implementation of the main help request and volunteer-related flows by replacing placeholder screens with backend-integrated functionality.

### Motivation

The Android app currently includes incomplete or placeholder flows for help requests and volunteer actions. These features should be usable, consistent with the existing Android architecture, and aligned with the current backend contract.

### Scope

This issue covers Android-only work for:
- Request Help
- Available to Help
- Assigned Request
-  My Help Requests. 

No backend or web changes are included.

### Feature Details

- Replace placeholder screens with real implementations
- Integrate with existing backend endpoints
- Support the required guest vs logged-in behavior for each screen
- Handle loading, empty, and error states consistently

### Acceptance Criteria

- [x] Code implemented
- [ ] Unit tests passed
- [ ] Documentation updated

### Notes

- All work should stay inside the Android project
- Existing Android navigation, auth/session, repository, and shared UI patterns should be reused
- Backend limitations must be respected
- Guest Request Help submission may still be limited if backend authentication is required

PR #195, opened 2026-04-03
feat: implement help flows (request, availability, assigned, my requests)

gulcetahtasiz:
This PR implements and integrates the core help-related flows on the Android side, replacing placeholder screens with backend-connected features.

Implemented:
- Request Help (single screen for guest + logged-in, with validation and profile prefills)
- Available to Help (toggle with backend integration, logged-in only)
- Assigned Request (backend-integrated, logged-in only)
- My Help Requests (guest + logged-in compatible)

Key details:
- Request Help uses one screen for both guest and logged-in users
- Guest users can fill the Request Help form, but submission is currently blocked by backend authentication requirements
- Logged-in users can submit real help requests
- Available to Help includes loading, success, and error handling
- Assigned Request shows the current active assignment from backend
- My Help Requests:
  - guest -> fallback message
  - logged-in -> real backend data

Technical notes:
- Followed existing Compose, navigation, and repository patterns
- Integrated real backend endpoints where supported by the current backend contract
- Implemented loading, empty, and error states across all screens

Limitations:
- POST /help-requests requires authentication, so guest submission is blocked by backend
- Assigned Request is based on the current active assignment endpoint, not assignment history
- The “single active request” rule is currently enforced on Android side only

Closes [#193](https://github.com/bounswe/bounswe2026group6/issues/193)

### DIFF
--- a/android/app/src/main/java/com/neph/features/myhelprequests/data/MyHelpRequestsRepository.kt
+++ b/android/app/src/main/java/com/neph/features/myhelprequests/data/MyHelpRequestsRepository.kt
@@ -7,6 +7,7 @@ import org.json.JSONObject
 
 data class MyHelpRequestUiModel(
     val id: String,
+    val guestAccessToken: String? = null,
     val helpTypes: List<String>,
     val helpTypeSummary: String,
     val description: String,
@@ -49,7 +50,12 @@ object MyHelpRequestsRepository {
         return buildList {
             for (trackedRequest in trackedRequests) {
                 val request = RequestHelpRepository.fetchGuestHelpRequest(trackedRequest) ?: continue
-                add(mapRequest(request))
+                add(
+                    mapRequest(
+                        request = request,
+                        guestAccessToken = trackedRequest.guestAccessToken
+                    )
+                )
             }
         }.distinctBy { it.id }
     }
@@ -65,7 +71,28 @@ object MyHelpRequestsRepository {
         return response.optJSONObject("request")?.let(::mapRequest)
     }
 
-    private fun mapRequest(request: JSONObject): MyHelpRequestUiModel {
+    suspend fun markGuestRequestAsResolved(
+        requestId: String,
+        guestAccessToken: String
+    ): MyHelpRequestUiModel? {
+        val response = JsonHttpClient.request(
+            path = "/help-requests/$requestId/status?guestAccessToken=$guestAccessToken",
+            method = "PATCH",
+            body = JSONObject().put("status", "RESOLVED")
+        )
+
+        return response.optJSONObject("request")?.let {
+            mapRequest(
+                request = it,
+                guestAccessToken = guestAccessToken
+            )
+        }
+    }
+
+    private fun mapRequest(
+        request: JSONObject,
+        guestAccessToken: String? = null
+    ): MyHelpRequestUiModel {
         val description = request.optString("description").trim()
         val helpTypes = request.optJSONArray("helpTypes").toStringList().map(::formatHelpType)
         val status = request.optString("status").ifBlank { "Unknown" }
@@ -78,6 +105,7 @@ object MyHelpRequestsRepository {
 
         return MyHelpRequestUiModel(
             id = request.optString("id"),
+            guestAccessToken = guestAccessToken,
             helpTypes = helpTypes,
             helpTypeSummary = buildHelpTypeSummary(helpTypes),
             description = description,
--- a/android/app/src/main/java/com/neph/features/myhelprequests/presentation/MyHelpRequestsScreen.kt
+++ b/android/app/src/main/java/com/neph/features/myhelprequests/presentation/MyHelpRequestsScreen.kt
@@ -12,7 +12,7 @@ import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.filled.Help
+import androidx.compose.material.icons.automirrored.filled.Help
 import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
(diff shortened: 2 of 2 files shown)


kckagancan (changes requested):
# NEPH Android Review — Major Issues Only

## `android/app/src/main/java/com/neph/features/assignedrequest/data/AssignedRequestRepository.kt`
- `fetchAssignedRequests()` effectively assumes **a single assignment** by reading one `assignment` object and wrapping it in `listOf(...)`. If the product/backend can return multiple assigned items, this is the wrong data model.
- `helpType` is stored as a **single string** (`need_type`). This does not align well with the newer **multi-help-type** request model.
- `locationLabel` is too weak: it only shows `Lat/Lon` or `"Location unavailable"`. For an assigned-response flow, this is not operationally strong enough if better location fields exist. (Location may use the same things as in profile)
- The data shape here is notably different from the help-request mapping elsewhere (`snake_case`, flat lat/lon, etc.), which raises a **contract consistency risk**.

## `android/app/src/main/java/com/neph/features/assignedrequest/presentation/AssignedRequestScreen.kt`
- The screen is visually built like a **list**, but the repository currently behaves like **single-assignment** logic. That mismatch is risky.
- The screen is too **passive** for an assigned-workflow page:
  - no detail action
  - no responder/requester coordination action
  - no stronger operational UI
- Status is shown as a **raw backend string**, which may surface ugly/internal enum values directly to users.
- Location display is still too weak because it depends on the repository’s lat/lon-only label.

## `android/app/src/main/java/com/neph/features/availability/data/AvailabilityRepository.kt`
- Availability state is primarily driven by **local cached state** (`SharedPreferences`) with no visible “fetch real backend state on app start/open” flow here. This can drift from server truth.
- The model is probably too thin for the feature’s importance: only `isAvailable` and `assignmentId` are tracked, which may become fragile for a richer volunteer assignment flow.
- The request body uses `isAvailable`; if backend expects a different naming convention, this is brittle. Combined with other files, the backend contract handling in this feature feels inconsistent.

## `android/app/src/main/java/com/neph/features/home/presentation/HomeScreen.kt`
- `Request Help` is opened directly with **no active-request gate**. This conflicts with the rule that a user should not proceed into a new request flow when they already have an active one.
- Availability state is loaded from the local repository state, which may not reflect the **real backend state**.
- If turning availability on results in an assignment, the UI only shows an info message; it does **not guide the user** into the assigned-request flow.
- `Available to Help` is shown to **all authenticated users** with no visible role/capability check.

## `android/app/src/main/java/com/neph/features/myhelprequests/data/MyHelpRequestsRepository.kt`
- The model still assumes a **single `helpType` string**, while the request creation UI allows **multiple help types**.
- Location display is too weak again: only lat/lon or `"Location unavailable"`, despite the request form collecting more meaningful location structure.
- This repository is shaped like a **generic history list**, not an **active-request-status-focused** model.
- Its backend field names/shapes differ quite a lot from the assigned-request side, which creates a broader **API contract consistency risk**.

## `android/app/src/main/java/com/neph/features/myhelprequests/presentation/MyHelpRequestsScreen.kt`
- This screen does **not match the intended product role**. It behaves like a passive list/history page instead of:
  - showing the user’s **current active request**
  - surfacing its **status clearly**
  - showing **assigned responder details**
  - allowing **cancel**
  - showing “no active request” when none exists
- There is **no detail flow**, **no cancel action**, and **no assigned-person info**.
- Status is shown raw and location is weak, so even the visible request summary is not very strong.
- Structurally this page is still closer to “request history list” than “active request tracking page”.

## `android/app/src/main/java/com/neph/features/requesthelp/data/RequestHelpRepository.kt`
- This is one of the **biggest issues** in the whole set.
- The submission model only contains:
  - `needType`
  - `description`
  - `isSavedLocally`
- That means the real structured form data is **not being sent structurally**:
  - affected people count
  - risk flags
  - vulnerable groups
  - blood type
  - location fields
  - contact info
  - confirmation
- `needType` is still a **single string**, even though the UI supports **multi-select help types**.
- The repository requires a token for creation, so the **guest flow is not real** at submission time.
- The active-request check is only implemented for authenticated users, which reinforces that the guest version of the flow is incomplete.

## `android/app/src/main/java/com/neph/features/requesthelp/presentation/RequestHelpScreen.kt`
- The form UI is close to the intended design, but the actual submission is still **old-contract based**:
  - most structured fields are flattened into the `description`
  - help types are joined into one string
- This makes the feature look rich in UI while remaining **structurally weak in data**.
- The guest flow is only cosmetic right now:
  - guests can fill the form
  - but cannot actually submit
- The “only one active request” rule is enforced **too late**:
  - the user can fill the whole form
  - only at submit time they are blocked
  - instead of being redirected to their current active request early
- Success handling is weak:
  - it sets a success message
  - then immediately navigates back
  - so the success state may never be meaningfully seen
- `Cancel` uses back navigation rather than a guaranteed **return-to-home** behavior.

## `android/app/src/main/java/com/neph/navigation/AppNavGraph.kt`
- The active-request rule is **not enforced at navigation level**:
  - Home always navigates directly to `RequestHelp`
  - there is no guard that redirects a user with an active request to the active-request/status page
- Request Help route is guest-accessible, but the real submit flow requires auth. That makes the guest flow **navigation-open but functionally incomplete**.
- This file confirms that the active-request gating problem is real and not being handled elsewhere.


---

# Highest-Priority Problems Across These Files
- **Request Help UI is rich, but the backend submission contract is still minimal and old-style.**
- **Multi-select help type exists in UI but is collapsed into a single string at submission/display layers.**
- **Guest Request Help flow is not truly supported.**
- **Single active-request rule is enforced too late and not at navigation/entry level.**
- **My Help Requests does not yet behave like the intended active-request status tracking page.**
- **Availability / Assigned Request flow still feels too shallow for the operational workflow it is supposed to support.**

gulcetahtasiz:
Updated Android to match the current backend and workflow expectations:

- Reworked Request Help to send the new structured backend payload instead of the old minimal contract.
- Moved the active-request guard earlier, so users with an existing active request are redirected before completing the form.
- Updated My Help Requests to parse the new response shape correctly and present the current active request more clearly.
- Improved Assigned Request and availability flow with stronger status/location handling and assignment actions.
- Reduced reliance on stale local availability state by refreshing assignment-related state from backend-supported flow.

Still intentionally unchanged:
- Guest submission is not enabled because backend still requires auth.
- User-side cancel / assigned responder details are still limited by missing backend support.

Could you recheck @kckagancan ?

Issue #197, opened 2026-04-05
[FIX] Add read endpoint for availability


berat-sayin:
To display the availability status on the Android side based on the backend truth, we need a read endpoint that returns the current availability state. Currently, only toggle and assignment endpoints exist; as a result, if a user is actually available but hasn't received an assignment yet, the status is not reflected correctly.

PR #198, opened 2026-04-05
fix: Add read endpoint for availability module and update its tests

berat-sayin:
Related to the issue [#197](https://github.com/bounswe/bounswe2026group6/issues/197).

🚀 Features
   - New Availability Status Endpoint: Added GET /api/availability/status to allow the Android client (and
     other consumers) to fetch the current backend state of a volunteer. 
       - Returns isAvailable (boolean), the volunteer record, and any active assignment.
       - This ensures the UI remains synced with the "source of truth" even if an assignment was triggered
         automatically on the backend.

  🛠 Technical Changes
   - Service Layer: Implemented getAvailabilityStatus in service.js to aggregate volunteer data and active
     assignments.
   - Controller Layer: Added handleGetAvailabilityStatus to manage the request/response cycle for the new
     status check.
   - Routes: Exposed the new functionality via GET /status within the availability router.

  🧪 Testing & Quality
   - Integration Tests: Added a new test case to availability.integration.test.js to verify the status
     endpoint behavior for both new and active volunteers.
   - Schema Alignment: Updated seedHelpRequest in the test suite to include mandatory fields
     (contact_full_name, contact_phone) required by the latest database schema, fixing previously failing
     tests.
   - Validation: Verified all 8 integration tests in the availability module are passing.

  📦 Impact
  This change resolves the desync issue where a volunteer's status might not reflect their actual
  availability or assignment state in the Android application until a manual toggle or sync occurred.

### DIFF
--- a/backend/src/modules/availability/controller.js
+++ b/backend/src/modules/availability/controller.js
@@ -24,7 +24,7 @@ async function handleSetAvailability(req, res) {
     return res.status(200).json(result);
   } catch (error) {
     console.error('Error in setAvailability:', error);
-    return res.status(500).json({ code: 'INTERNAL_ERROR', message: error.message });
+    return res.status(500).json({ code: 'INTERNAL_ERROR', message: 'An internal server error occurred' });
   }
 }
 
@@ -39,7 +39,7 @@ async function handleSyncAvailability(req, res) {
     return res.status(200).json(result);
   } catch (error) {
     console.error('Error in syncAvailability:', error);
-    return res.status(500).json({ code: 'INTERNAL_ERROR', message: error.message });
+    return res.status(500).json({ code: 'INTERNAL_ERROR', message: 'An internal server error occurred' });
   }
 }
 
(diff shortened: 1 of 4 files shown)


AlperKartkaya (changes requested):
Good & focused change, solving the client-server sync gap from #197.

**Must-do (before or right after merge):**
- Add unit tests for `getAvailabilityStatus` (service) and `handleGetAvailabilityStatus` (controller), as every other function in both files has unit test coverage and this breaks the pattern.

**Nits:**
- `infra/dcompose/package-lock.json` looks unintentional (empty lockfile with no deps). Consider removing.
- The `error.message` exposure in the 500 handler is consistent with the existing code, but worth a module-wide follow-up to return generic messages to clients.

Otherwise: code is clean, route ordering is correct, integration test is solid, and the `seedHelpRequest` schema fix was needed.

berat-sayin:
### Changes made:
1. Removed Unintentional File: Deleted infra/dcompose/package-lock.json as it was an empty lockfile.
   
2. Added Unit Tests: 
       * Added tests for getAvailabilityStatus in backend/tests/unit/modules/availability/service.test.js.
       * Added tests for handleGetAvailabilityStatus in
         backend/tests/unit/modules/availability/controller.test.js.
   
3. Improved Error Handling: Updated all 500 error handlers in the availability controller
      (backend/src/modules/availability/controller.js) to return a generic "An internal server error
      occurred" message instead of exposing the raw error.message.

PR #220, opened 2026-04-06
fix: Auto-assign when new help request created

berat-sayin:
Related to [#225](https://github.com/bounswe/bounswe2026group6/issues/225)

Changes:
   - backend/src/modules/availability/repository.js: Added findMatchingVolunteerForRequest to find an
     available volunteer who doesn't have an active assignment and matches the request's need type.
   - backend/src/modules/availability/service.js: Added and exported tryToAssignRequest to encapsulate the
     assignment logic (creating an assignment record and updating the request status to ASSIGNED).
   - backend/src/modules/help-requests/service.js: Updated createMyHelpRequest to call tryToAssignRequest
     immediately after a request is successfully created. It now returns the updated request (which will
     have MATCHED status if a volunteer was found).
   - Tests:
     - Added a new integration test case in
       backend/tests/integration/modules/help-requests/help-requests.integration.test.js to verify that a
       new request is auto-assigned to an already available volunteer.
     - Updated unit tests in both availability and help-requests modules to reflect the new functionality
       and ensure correct mocking.

### DIFF
--- a/backend/src/modules/availability/repository.js
+++ b/backend/src/modules/availability/repository.js
@@ -102,6 +102,32 @@ async function findMatchingRequestForVolunteer(volunteerId) {
   return result.rows[0] || null;
 }
 
+async function findMatchingVolunteerForRequest(requestId) {
+  const requestSql = `SELECT need_type FROM help_requests WHERE request_id = $1;`;
+  const rResult = await query(requestSql, [requestId]);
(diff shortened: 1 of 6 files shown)


mehmetcangurbuz08 (approved):
Looks solid. Merging.

PR #260, opened 2026-04-14
feat: standardize DB migrations under backend and add local/prod migration workflow

### LINKED ISSUE #259: [DevOps] Implement workflow for DB migration to the prod
Write a CI pipeline action for DB migration actions to update the prod db. It has to run when the db files change while merging the main branch.

mehmetcangurbuz08:
Unifies migration flow so local and production stay in sync.

Changes
Uses only backend/migrations for schema changes.
Adds idempotent migration runner (schema_migrations tracking).
Runs migrations on backend startup in Docker.
Adds production workflow to apply migrations on main.

Closes #259

### DIFF
--- a/.github/workflows/db-prod-sync.yml
+++ b/.github/workflows/db-prod-sync.yml
@@ -57,18 +57,17 @@ jobs:
           echo "migration_changed=$MIGRATION_CHANGED" >> "$GITHUB_OUTPUT"
 
       - name: Install PostgreSQL client
-        if: steps.changes.outputs.init_sql_changed == 'true' || steps.changes.outputs.migration_changed == 'true'
+        if: steps.changes.outputs.init_sql_changed == 'true'
         run: |
           sudo apt-get update
           sudo apt-get install -y postgresql-client
 
-      - name: Apply production DB updates
-        if: steps.changes.outputs.init_sql_changed == 'true' || steps.changes.outputs.migration_changed == 'true'
+      - name: Apply init.sql on production (guarded)
+        if: steps.changes.outputs.init_sql_changed == 'true'
         env:
           PROD_DATABASE_URL: ${{ secrets.PROD_DATABASE_URL }}
           ALLOW_PROD_SCHEMA_RESET: ${{ secrets.ALLOW_PROD_SCHEMA_RESET }}
           INIT_SQL_CHANGED: ${{ steps.changes.outputs.init_sql_changed }}
-          MIGRATION_CHANGED: ${{ steps.changes.outputs.migration_changed }}
         shell: bash
         run: |
           set -euo pipefail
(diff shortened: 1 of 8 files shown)


kckagancan (dismissed):
# PR Review — Migration Flow / DB Sync

## Overall assessment

The goal of this PR is good: centralize schema changes and reduce drift between local and production.  
However, in its current form, I do **not** think this is production-safe yet.

### Main blocking concerns

1. **There is no truly unified migration flow**
   - Local/backend startup uses `backend/scripts/apply-migrations.js`
   - Production uses a separate GitHub Actions + bash + `psql` flow
   - So local and production are **not** using the same migration implementation

2. **The `check -> apply -> markApplied` flow is not atomic**
   - A migration can be applied successfully
   - But recording it in `schema_migrations` can fail afterward
   - In that case, the system may try to re-apply an already executed migration later

3. **There is no DB-level locking**
   - GitHub Actions concurrency only protects the workflow itself
   - It does not protect against concurrent execution from app startup or any other process
   - This creates a race condition risk

4. **If startup migrations and production workflow migrations both exist, they can conflict**
   - Production should have exactly one migration execution path

---

## 1) `.github/workflows/db-prod-sync.yml`

### Critical findings

#### 1. Migration execution and migration tracking are not atomic
The workflow applies the SQL migration first, then records it in `schema_migrations` in a separate command.

Risk scenario:
- the migration SQL succeeds
- the `INSERT INTO schema_migrations` step fails
- the database has already changed
- the next run thinks the migration was never applied and tries again

This is one of the most serious issues in the current design.

#### 2. No DB-level lock
Workflow-level concurrency is helpful, but not sufficient.

If production app startup also runs migrations, or if another process runs them, two executors can:
- both see that a migration is not yet recorded
- both try to apply it
- collide during execution

This is a classic race condition.

#### 3. Allowing `init.sql` to run on production is dangerous
Even with `ALLOW_PROD_SCHEMA_RESET=true`, this is still risky by design.

`init.sql` is usually for:
- bootstrap
- fresh database initialization
- local setup

Production schema evolution should ideally rely only on versioned migrations, not a reset-style SQL path.

#### 4. File path is interpolated directly into SQL in the workflow
The workflow builds SQL strings using the file name directly.  
Even if file names are expected to stay simple, this is still not a safe pattern and can break on unexpected characters.

### Recommendation
- Use a single migration execution path for production
- Add a DB-level lock such as a Postgres advisory lock
- Make migration apply + migration record update failure-safe / atomic
- Remove `init.sql` from the production schema update path

---

## 2) `backend/migrations/README.md`

### Critical finding

#### Startup migration behavior is unclear for production
The README states that the backend container runs migrations on startup.

But production also has a separate workflow that applies migrations.

If both are intended to be active in production, that is a critical risk:
- the same migration can be triggered from two different places
- race conditions become possible
- production deploys become unstable

### Recommendation
The README should explicitly define one rule, for example:

- **Either:** startup migrations are for local/dev only
- **Or:** production migrations are only applied by CI/CD

This needs to be unambiguous.

---

## 3) `backend/scripts/apply-migrations.js`

### Critical findings

#### 1. `isAlreadyApplied -> applyMigration -> markApplied` is not atomic
The script:
- checks whether the migration was already applied
- runs the migration
- then records it in `schema_migrations`

There is no guarantee that these steps stay consistent as a unit.

Risk scenario:
- migration SQL succeeds
- `markApplied()` fails
- the migration is already in the database
- but it is not recorded
- the next run tries to execute it again

This is a blocking issue.

#### 2. Race condition exists here as well
If two processes run this script at the same time:
- both can observe that the migration is not yet applied
- both can try to apply it

The `schema_migrations` table alone does not prevent this.  
The issue is the race between the check and the actual execution.

#### 3. The comment about idempotency is too optimistic
The code comment says:

> Idempotency is managed by schema_migrations table.

That is not fully true in the current design.  
The table provides basic tracking, but it does **not** solve:
- concurrent execution
- partial failure after apply
- re-apply attempts after tracking failure

### Important note
This file is better than the workflow in one respect: it uses parameterized queries for checking and inserting migration records.  
That part is good.

However, the main critical problems remain:
- no locking
- no atomic apply/mark behavior

### Recommendation
- Add a DB-level lock
- Make migration execution + tracking failure-safe
- Reuse the same migration runner logic in both local and production instead of maintaining two different implementations

---

## 4) `docker-compose.yml`

### Critical findings

#### 1. Backend startup runs migrations
```yaml
command: ["sh", "-c", "npm run start:with-migrations"]

For local development, this is understandable.
But if production also has a dedicated workflow that applies migrations, then this becomes a serious conflict point.

Risk scenario:

CI/CD workflow starts applying migrations
backend container restarts at the same time
startup migration also starts
both attempt to run the same schema changes
2. init.sql only runs for a fresh Postgres volume

The docker-entrypoint-initdb.d mechanism runs only when the data directory is initialized for the first time.

That means:

a developer with a fresh volume gets the new init.sql
a developer with an existing volume does not

So the claim that local and production stay in sync is incomplete in practice.
Even local environments may diverge depending on volume history.

Recommendation
Do not combine startup migrations and separate production workflow migrations unless there is a strict coordination/locking strategy
Keep init.sql as bootstrap-only
Use versioned migrations for schema evolution
Final conclusion

I agree with the intention of the PR, but I do not think it is safe to merge as-is for production use.

The two main blockers are:
Production does not have a single migration execution path
The migration runner is not atomic and not concurrency-safe
Most serious failure modes
the same migration being applied concurrently by different processes
a migration being applied successfully but not recorded, causing a later re-apply attempt
Suggested direction before merge
Choose exactly one production migration path
Add Postgres advisory locking
Make apply + markApplied failure-safe
Remove init.sql from production schema update flow
Avoid maintaining separate migration logic for local and production

mehmetcangurbuz08:
Nice catches. I fix the important ones. Can you review it again? @kckagancan

kckagancan (approved):
Nice fixes. Ready to merge now.

PR #310, opened 2026-04-19
feat: Add gathering areas backend module with Overpass integration, GeoJSON response, and integration test coverage

### LINKED ISSUE #309: [FEATURE] [Location] Implement nearby gathering areas endpoint with Overpass integration
We need a backend endpoint that returns nearby disaster gathering and shelter points based on user coordinates.
This issue covers Overpass integration, response standardization, error handling, and integration test coverage.

Scope

Add GET /api/gathering-areas/nearby with lat, lon, radius, and limit query params
Query Overpass for emergency=assembly_point and amenity=shelter
Return data as GeoJSON FeatureCollection
Validate input parameters (lat, lon, radius, limit)
Map provider timeout and unavailable cases to standard API errors
Add in-memory cache for repeated queries
Update backend README with endpoint details and sample response
Add integration tests for success and failure scenarios
Acceptance Criteria

Endpoint returns 200 with valid GeoJSON response for valid input
Invalid query params return 400
Provider timeout returns 504
Provider unavailable or invalid payload returns 503
Cache works for identical requests
Integration tests pass
README includes endpoint details and sample response

mehmetcangurbuz08:
This PR adds backend support for fetching nearby disaster gathering and shelter points based on user coordinates.

What is included

New endpoint: GET /api/gathering-areas/nearby with lat, lon, radius, and limit query params
New gathering-areas backend module with routes, controller, service, and validators
Overpass integration for emergency=assembly_point and amenity=shelter data
GeoJSON FeatureCollection response format for frontend map/list usage
Distance calculation and sorting by nearest result
Deduplication and malformed element filtering
Input validation for lat, lon, radius, and limit
Error mapping for provider unavailable and timeout scenarios
In-memory cache for repeated nearby queries
Router integration under /api/gathering-areas
Backend README updates with endpoint details and sample response
Testing

Added integration tests for success flow, validation, default values, max clamp behavior, cache behavior, sorting, deduplication, and malformed data handling
Added provider failure tests for unavailable, timeout, network errors, and invalid payload shapes
All related gathering-areas integration tests pass locally
Notes

No database migration is required for this change
This PR is focused on backend API readiness and stable contract for frontend integration

Closes #309

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -12,6 +12,8 @@ Main feature areas:
 - `src/modules/profiles` — user profile, privacy, health, and location data
 - `src/modules/help-requests` — request creation, tracking, and status updates
 - `src/modules/availability` — volunteer availability, matching, and assignment-related flows
+- `src/modules/location` — country tree, geocoding search, and reverse geocoding support
+- `src/modules/gathering-areas` — nearby disaster assembly and shelter points via Overpass
 
 ## Structure
 
@@ -73,7 +75,7 @@ GET http://localhost:3000/health
 
 ## Important endpoints
 
-The backend exposes these top-level paths:
+The backend exposes these top-level module paths (concrete endpoints are listed below):
 
 - `GET /health`
 - `GET /api`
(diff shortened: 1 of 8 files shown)


erinc00 (changes requested):
I checked the PR and the overall direction looks good, but I do not think it is ready to merge yet. The endpoint design, GeoJSON response shape, validation, and test coverage are all useful additions. However, I think the following points should be addressed before approval:

backend/src/modules/gathering-areas/service.js

The cache key currently uses lat/lon.toFixed(6), which is quite precise for a nearby-query cache and may reduce cache hit rate significantly under normal GPS jitter. This is not necessarily wrong, but I think it would be worth checking whether a slightly coarser precision or a deliberate bucketing strategy would give more practical reuse for mobile map/list flows.


Also, a couple of non-blocking points may be worth keeping in mind for later:

backend/src/modules/gathering-areas/controller.js

OVERPASS_INVALID_PAYLOAD is currently mapped to the same outward response as OVERPASS_UNAVAILABLE. That may be acceptable from a client-facing simplicity standpoint, but these are operationally different failure modes. This is probably okay for now, but it may be worth making this distinction more explicit in comments/logging later, since invalid payloads can indicate provider contract drift rather than simple downtime.

backend/src/modules/gathering-areas/service.js

The cache currently stores and returns the same object reference. That works as long as nothing mutates the result after retrieval, but it does make the cache vulnerable to accidental shared-state bugs later. This is probably fine for now, but if we expect this contract to remain stable over time, it may be safer to freeze or clone cached results later.

mehmetcangurbuz08:
Nice catch, thanks for calling this out. I fixed the mentioned issues. Can you review it again?

erinc00 (approved):
Looks good to me now. The main issue around cache-key precision has been addressed, and I also like that it is now backed by a small-GPS-jitter cache reuse test. The overall endpoint shape, validation, sorting, filtering, provider error handling, and test coverage look solid. Approving and merging.

PR #312, opened 2026-04-19
feat(web): add reusable location map infrastructure for shared location flows

erinc00:
This PR introduces a reusable location/map foundation on the web side, so upcoming location-based screens can build on shared components instead of one-off implementations.

What was added:

- Reusable map components:
  - Leaflet-based map container
  - Click-to-place pin
  - Draggable marker to update coordinates
  - SSR-safe dynamic map wrapper
- Reusable LocationPicker:
  - Search integration (/api/location/search)
  - Reverse geocode integration (/api/location/reverse)
  - Current location support (navigator.geolocation)
  - Shared picker value model for downstream forms/screens
- Shared location types:
  - Tree/search/reverse response types
  - Administrative + coordinate data structures
- Location API helpers:
  - Tree, search, and reverse helper methods
- Web dependencies:
  - leaflet
  - react-leaflet
  - @types/leaflet
 
Scope and non-goals:

- No backend changes in this PR.
- No direct integration yet into existing profile/complete-profile flows.
- Test-only preview route was intentionally excluded from commit/PR.

Validation:

- npm run build completed successfully
- TypeScript checks and static route generation passed

Files included:

- package.json
- package-lock.json
- location.ts
- location.ts
- LeafletLocationMap.tsx
- LocationPickerMap.tsx
- LocationPicker.tsx
- index.ts

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -10,7 +10,6 @@
         "next": "^16.2.0",
         "react": "^19.2.4",
         "react-dom": "^19.2.4",
-        "react-leaflet": "^5.0.0",
         "tailwind-merge": "^3.5.0"
       },
       "devDependencies": {
@@ -696,17 +695,6 @@
       ],
       "engines": {
         "node": ">= 10"
-      }
-    },
-    "node_modules/@react-leaflet/core": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/@react-leaflet/core/-/core-3.0.0.tgz",
-      "integrity": "sha512-3EWmekh4Nz+pGcr+xjf0KNyYfC3U2JjnkWsh0zcqaexYqmmB5ZhH37kz41JXGmKzpaMZCnPofBBm64i+YrEvGQ==",
-      "license": "Hippocratic-2.1",
-      "peerDependencies": {
-        "leaflet": "^1.9.0",
-        "react": "^19.0.0",
-        "react-dom": "^19.0.0"
       }
     },
     "node_modules/@swc/helpers": {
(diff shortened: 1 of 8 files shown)


kckagancan (changes requested):
Looks good overall, but there are two issues:

Medium: search input is not semantically labeled (accessibility gap).
Evidence: [LocationPicker.tsx:176], [LocationPicker.tsx:179-184] [TextInput.tsx:15-19]
Why this matters: the visible text is rendered as helper text, not an input label, so accessible name falls back to placeholder.

Medium: reverse-geocode failure path emits a partially empty value object, which can produce inconsistent downstream form state.
Evidence: [LocationPicker.tsx:117-129]
Why this matters: placeId and displayName are blank while coordinates are present, which can look selected but fail later validations/assumptions.

erinc00:
Thanks for the review. I pushed the fixes based on your comments. Could you please check it again @kckagancan ?

kckagancan (changes requested):
Thanks for the fixes @erinc00 , I can see the updates were applied.

There is still one remaining issue to address:

Map lifecycle regression: the Leaflet map can be re-initialized on normal parent re-renders because the initialization effect depends on changing props (including callback identity), and the cleanup removes/recreates the map instance.
[LeafletLocationMap.tsx:48]
[LeafletLocationMap.tsx:73]
[LeafletLocationMap.tsx:79]
[LocationPicker.tsx:234]
Please make map initialization stable (single mount) and update center/marker via separate effects without tearing down the map instance on routine renders.

erinc00:
I fixed based on your comments. Can you check it again @kckagancan ?

kckagancan (approved):
Great work. Everything is fixed now. Merging.

PR #316, opened 2026-04-19
feat(web): integrate hybrid location picker into profile flows

erinc00:
This PR integrates the shared hybrid location selection flow into profile-related web forms by combining:

- Map/search-based location picking
- Backend-driven location tree dropdowns
- Compatibility-safe location payload mapping for existing backend fields

Changes
- Refactored complete profile form to use shared location infrastructure.
  - CompleteProfileForm.tsx
- Refactored profile edit view to use the same hybrid location flow.
  - ProfileView.tsx
 - Added shared location-tree helper utilities for key/label mapping and address parsing.
   - locationTree.ts

### DIFF
--- a/web/src/components/feature/auth/CompleteProfileForm.tsx
+++ b/web/src/components/feature/auth/CompleteProfileForm.tsx
@@ -10,10 +10,19 @@ import { Checkbox } from "@/components/ui/selection/Checkbox";
 import { ProfileInfoRow } from "../../ui/display/ProfileInfoRow";
 import { SaveActionBar } from "../../ui/display/SaveActionBar";
 import { HelperText } from "@/components/ui/display/HelperText";
+import { LocationPicker, LocationPickerValue } from "@/components/feature/location";
 import { bloodTypeOptions } from "@/lib/bloodTypes";
 import { countryCodeOptions } from "@/lib/countryCodes";
 import { expertiseOptions, professionOptions } from "@/lib/profileOptions";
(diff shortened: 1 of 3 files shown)


kckagancan (changes requested):
Nice refactor overall @erinc00  — consolidating the location logic into shared helpers and reusing the same hybrid flow across complete-profile and profile edit makes sense. I still see two product-level issues before approval though:

Medium: Location tree loading is now part of the critical path for profile-related screens, so a failure in `/location/tree` can block flows that previously did not depend on it.
[CompleteProfileForm.tsx]
[ProfileView.tsx]
Expected fix:
Keep profile data / onboarding usable even if the location tree request fails. The location tree should degrade the location-editing experience, not break the whole screen or make profile completion impossible.

Medium: The new label-to-key mapping is too strict, which can fail to reconstruct already-saved locations when backend labels differ even slightly from the tree labels.
[ProfileView.tsx]
[locationTree.ts]
Expected fix:
Make the country/city/district matching more robust (for example by normalizing labels consistently before comparison), so existing persisted location data can still be resolved into dropdown state reliably.

NOTE: E2E test is failed due to a bug within the test code itself. Don't mind that fail.

erinc00:
I fixed the issues you have mentioned. Can you review again? @kckagancan

kckagancan (approved):
“Rechecked after the latest commit. The two issues I raised appear to be addressed now. Great work. Merging.

Issue #288, opened 2026-04-16
[FEATURE] [Location] Add manual location selection


erinc00:
**Description**
Users should be able to manually select a location in the application. The selected location should be stored and reused where relevant.

**Tasks**
- Design web UI for manual location selection
- Design mobile UI for manual location selection
- Implement backend support for saving selected location
- Implement retrieval/update flow for selected location
- Define how the selected location is shown back to the user
- Handle invalid/empty selection cases

**Acceptance Criteria**
- User can manually select a location
- Selected location can be saved and retrieved
- Selected location can be reused in related flows

**Dependencies**
- Shared map and location infrastructure

PR #318, opened 2026-04-20
feature: implemented complete location selection save/retrieve flow

kckagancan:
## Summary

This PR completes manual location selection end-to-end for Android profile location data.

Users can now select a location manually, have it stored persistently through the backend, and receive the same structured location data back when the profile is loaded again.

## What Changed

- Android now sends and reads structured profile location fields more reliably
- Backend now persists and returns structured administrative location data for profiles
- Tests were updated on both Android and backend to cover save/retrieve roundtrip behavior

## Validation

- `npm run test:integration`
- `./gradlew :app:test`

---

Completes the android part of the issue #288


kckagancan:
I now fully integrated android/ with existing backend/. Previous commit did not include the integration changes. Can you guys review all the changes now? @erinc00 @mehmetcangurbuz08

erinc00 (changes requested):
Thanks for the update. I reviewed the latest changes carefully. The overall direction looks good, but I still see a few important points that should be addressed or clarified before I would feel comfortable approving this.

## Blocking Findings

### 1. Fake backend semantics still do not look fully aligned with the real backend
`android/app/src/androidTest/java/com/neph/e2e/FakeNephBackend.kt`

The fake backend patch flow for location still appears to differ from the real backend’s partial update semantics, especially around overwrite/fallback behavior for fields like `city` and `address`.

That is risky because instrumentation/E2E tests can stay green while production behavior differs. Since this PR is specifically claiming end-to-end manual location persistence/retrieval reliability, I think the fake backend needs to mirror the real backend contract much more closely.

### 2. `locationProfile.coordinate` nullability contract needs to be explicitly preserved or redefined
`backend/src/modules/profiles/service.js`

This change appears to affect the shape/contract of `locationProfile.coordinate`. Previously, the contract looked like:

- if lat/lon are missing → `coordinate: null`
- if lat/lon are present → `coordinate: { ... }`

If that behavior has changed, this can become a backward compatibility issue for clients that rely on `null` rather than a partially filled object. I think this contract needs to be made explicit and locked down with tests.

### 3. `countryCode` derivation on Android is not fully reliable
`android/app/src/main/java/com/neph/features/profile/data/ProfileRepository.kt`

The Android side currently derives backend country code using `uppercase()` from the selected country value. That is only safe if the input is guaranteed to be a real country key / ISO-style code.

If a label ever reaches this point instead of a key, this could produce invalid values like `TURKEY` instead of `TR`. I think `countryCode` should only be set when it comes from a trusted key/code source; otherwise it should remain `null`.

## High Priority Findings

### 4. Lowercasing neighborhood `value` may break roundtrip assumptions
`android/app/src/main/java/com/neph/features/profile/data/LocationTreeRepository.kt`

`parseNeighborhoods()` lowercases the backend-provided `value` field. That may be fine if the contract explicitly defines these values as normalized case-insensitive slugs.

But if backend values are intended to be preserved exactly, this introduces a subtle roundtrip risk. I think the original value should be preserved, and case-tolerant matching should happen at comparison time if needed.

## Medium Priority Findings

### 5. Migration backfill scope is limited
`backend/migrations/20260420_120000__extend_location_profiles_administrative_fields.sql`

The migration adds the new structured location columns and backfills `display_address`, but it does not backfill fields like:

- `country_code`
- `district`
- `neighborhood`
- `extra_address`

This may be an intentional choice, which is okay. But if so, I think it should be called out explicitly. Otherwise, a best-effort backfill strategy for older records may be worth considering.

### 6. Test coverage is still a bit thin for the “roundtrip complete” claim
The current tests are a good start, but I think a few critical scenarios are still missing:

- preserving fields that are omitted in a PATCH
- explicit `null` behavior for clearing fields
- `coordinate` nullability contract

Given the PR summary, I think these should be covered to make the roundtrip guarantee more convincing.

Because of the first three items above, I would still keep this as **request changes** rather than approve at this stage.

kckagancan:
Thanks for the review. I fixed the issues you have stated, can you re-review please? @erinc00

erinc00 (changes requested):
Thanks for the update. I rechecked the latest version and there are still a few points I think we should address before merging.


### High

1. country_code can still store non-ISO values (data quality / contract drift risk)  
repository.js  
`countryCode` is uppercased but not validated as ISO-3166 alpha-2. Clients can still send values like `TURKEY`, and backend will persist them. This can create cross-client inconsistencies and weakens the API contract.  
Suggestion: validate `administrative.countryCode` as a strict 2-letter code (or known country set), otherwise return 400 or coerce to null with an explicit policy.

### Medium

1. Migration backfill is limited to display_address  
backend/migrations/20260420_120000__extend_location_profiles_administrative_fields.sql  
The migration adds multiple columns, but only `display_address` is backfilled. Existing rows will keep `country_code`, `district`, `neighborhood`, `extra_address`, `postal_code`, and `place_id` as null.  
Suggestion: either add best-effort backfill for legacy data or document this as an intentional decision in migration/PR notes.

2. Test setup executes all .sql files recursively under migrations  
global-setup.js  
The setup currently runs every `.sql` file found recursively. Future helper/example SQL files may be executed unintentionally and break tests.  
Suggestion: filter files by migration naming pattern (for example `YYYYMMDD_HHMMSS__*.sql`) before execution.

3. Android test reset does not appear to reset location-tree cache metadata  
Files: LocationData.kt, android/app/src/main/java/com/neph/features/profile/data/LocationTreeRepository.kt  
`resetLocationDataForTesting` resets the map, but repository cache state (`lastSyncedCountryCode`, `lastSyncedAtMs`) is separate and may leak across tests.  
Suggestion: add a test-only reset for repository cache metadata, or provide one deterministic reset utility used by tests.

kckagancan:
Thanks for the review again. I fixed the issues you have stated, can you re-review please? @erinc00

kckagancan:
The failing check is unrelated to this PR’s Android/backend changes.

`Backend CI` passes, and the only failing workflow is the web Playwright E2E suite. The failure comes from the existing onboarding test still trying to select a hardcoded neighborhood option (`bostanci`) that is no longer available in the current web location flow.

So this is a pre-existing / separate web test issue, not a regression introduced by this PR. This PR should still be mergeable once that unrelated E2E failure is taken into account.

erinc00 (approved):
Thanks for your effort, looks good now. Merging

PR #321, opened 2026-04-20
test(web): make onboarding neighborhood selection dynamic

erinc00:
Removed the hardcoded bostanci neighborhood selection in onboarding E2E.
After district selection, the test now picks the first available non-empty neighborhood option dynamically.
This aligns the test with the current backend-driven location tree flow and prevents timeout on stale option values.


kckagancan (changes requested):
The hardcoded neighborhood selection issue looks fixed, and the test now correctly moves past that step.

However, there are still some issues because the PR is still red. The onboarding E2E now fails later on a persisted city assertion (`expected "Istanbul", received "İSTANBUL"`), so the overall test flow is not fully resolved yet.

erinc00:
Can you merge now if everything is ok @kckagancan ?

kckagancan (approved):
Looks good now. Merging.

PR #323, opened 2026-04-20
feat(web): persist current location coordinates in profile flows

erinc00:
- Added coordinate persistence to web profile location save flows.
- Extended location picker output to carry selection metadata (source and accuracy) for current device, map pin, and search selections.
- Included latitude, longitude, coordinate, and administrative fields in location patch requests from both complete-profile and profile-edit forms.
- Expanded web location patch payload typing to align with backend location contract.
- Replaced hardcoded country code with dynamically resolved country code.


mehmetcangurbuz08 (approved):
Looks good overall. Well done. Merging.

PR #322, opened 2026-04-20
feat: implement permission-aware current location sharing for android

kckagancan:
This PR adds Android support for Share Current Location in profile flows.

What’s included:

requests runtime location permission when users enable Share Current Location
captures current (or last-known) device coordinates on save
sends coordinates with source and capturedAt through existing profile location APIs
handles permission denied and location unavailable cases safely without blocking profile save
Scope:

Android only
no web changes
no backend changes (reuses existing backend location/privacy infrastructure)
Validation:

Android unit tests updated for location payload behavior
Android build/tests pass for the project’s e2e test variant

This PR completes the android side of the issue #290

PR #324, opened 2026-04-21
fix: clear shared coordinates on disable and align location-permission save flow

kckagancan:
This PR tightens Android current-location sharing behavior.

- explicitly clears stored coordinates when Share Current Location is disabled
- makes save flow consistent when permission is revoked or location is unavailable by turning sharing off and clearing coordinates
- ensures users see completion feedback in Complete Profile (toast before navigation)

adds minimal payload tests for:

- coordinates included when sharing is enabled and location is available
- coordinates cleared when sharing is disabled
- no regression for unavailable-location payload behavior

This is a fix for PR  #322 
Can one of you guys review? @erinc00 @mehmetcangurbuz08


erinc00 (approved):
Looks good, merging

Issue #269, opened 2026-04-16
[FEATURE][Dashboard] Build aggregate emergency overview


mehmetcangurbuz08:
**Description**
The system should provide an aggregate emergency overview instead of showing only individual emergency items. The goal is to help admins and coordinators quickly understand the overall state of emergencies in the system.

**Tasks**
- Define which aggregate emergency metrics should be shown in the overview
- Design a dashboard section or overview page for aggregate emergency visibility
- Implement backend support for summary counts and aggregated emergency statistics
- Display active emergency count
- Display resolved/closed emergency count
- Display pending / in-progress / resolved breakdown
- Display urgency breakdown
- Display a recent activity summary
- Add optional region-based summary if feasible
- Handle loading, empty, and error states

**Acceptance Criteria**
- Admin users can view an aggregate emergency overview
- The overview includes meaningful high-level emergency counts and status breakdowns
- Aggregate information is understandable at a glance
- Loading, empty, and error states are handled properly

**Dependencies**
- Web-only admin panel
- Backend readiness for emergency aggregation

PR #329, opened 2026-04-21
feat: add aggregate emergency overview API and split admin endpoints into dedicated module

mehmetcangurbuz08:
This PR introduces backend support for the Admin Dashboard emergency overview and refactors admin routes out of the auth module.

What’s included
Added new admin-only endpoint: GET /api/admin/emergency-overview
Implemented aggregate metrics for:
active / resolved / closed emergency counts
status breakdown (pending, in-progress, resolved, cancelled)
urgency breakdown (low, medium, high)
recent activity summary (last 24h / 7d)
optional city-based region summary
Moved admin functionality from auth into a dedicated admin module:
repository, service, controller, routes
Registered admin module under /api/admin
Access control
Endpoint is protected with existing requireAuth + requireAdmin
Verified behavior for:
401 unauthenticated
403 non-admin
200 admin
Tests
Added/updated unit tests for new admin service/controller
Added integration test for /api/admin/emergency-overview
All related tests pass locally

Related to backend part of #269


erinc00 (changes requested):
Thanks for the refactor and for separating admin functionality into its own module. The direction makes sense, and the new emergency overview endpoint is useful for the dashboard. I went through the latest diff carefully, and I still see a few issues that I think should be addressed before merge.

## Review Findings

### Blocking

1. **Backward compatibility break for existing admin endpoints is not handled**  
   **Files:** `backend/src/modules/auth/routes.js`, `backend/src/routes/index.js`

   The old admin endpoints under `/api/auth/admin/*` are removed and replaced with `/api/admin/*`.

   That refactor is reasonable internally, but this is still an externally visible API path change. If any existing frontend/admin client, documentation, test environment, or staging integration still calls the old routes, this PR introduces a breaking change.

   I do not see any compatibility layer, deprecation note, redirect strategy, or evidence in this PR that all consumers were migrated together.
  
2. **`cancelledLast24Hours` / `cancelledLast7Days` are based on `created_at`, not cancellation time**  
   **File:** `backend/src/modules/admin/repository.js`

   In `recentActivity`, cancelled metrics are computed like this:

   - `status = 'CANCELLED'`
   - `created_at >= NOW() - INTERVAL ...`

   This does **not** measure recent cancellations. It measures requests that were **created recently and currently have status CANCELLED**.

   That can produce misleading dashboard numbers:
   - a request created 20 days ago but cancelled today will **not** be counted,
   - a request created yesterday and cancelled much later could still be counted depending on current state.

erinc00:
A few additional issues also stood out to me while reviewing the latest version. In addition to the earlier comments, I think the following points should also be addressed:

### High

1. **Region summary can overcount if `request_locations` is not strictly one-to-one with requests**  
   **File:** `backend/src/modules/admin/repository.js`

   The query currently does:

         ```sql
         FROM help_requests hr
         LEFT JOIN request_locations rl ON rl.request_id = hr.request_id
         GROUP BY city```
   
If `request_locations` is guaranteed one row per request, this is fine. But if the table can contain multiple rows per request (historical entries, duplicates, future extensions, migration leftovers, etc.), then the counts here will be inflated.

The current integration test only seeds one location row per request, so it would not catch this.

2. “Optional city-based region summary” is not actually optional in the API
**Files:** `backend/src/modules/admin/controller.js`, `backend/src/modules/admin/routes.js`, `backend/src/modules/admin/repository.js`

The PR summary says the endpoint includes an optional city-based region summary, but the implementation always computes and returns `regionSummary`, and there is no query parameter or toggle for it.

This is not a runtime bug, but it is a contract mismatch between the PR description and the API behavior.

erinc00 (approved):
Looks good to me overall. The main concerns from the previous revision appear to be addressed, especially making `regionSummary` actually optional, reducing the region-summary overcount risk, and introducing `cancelled_at` for more accurate cancellation activity metrics.

I still see a few smaller robustness / maintainability points (for example explicit null ordering in the lateral location selection, logging in controller error paths, and documenting/backfilling historical `cancelled_at` behavior), but I do not see a blocking issue at this stage.

Approved. Merging

PR #331, opened 2026-04-21
feat(web): implement permission-aware device location sharing flow

### LINKED ISSUE #290: [FEATURE] [Location] Add share current location flow
**Description**
Users should be able to share their current device location. The backend should accept and store the latest shared location where relevant, while handling permission and unavailable-location cases properly.

**Tasks**
- Implement permission-aware current location retrieval on web
- Implement permission-aware current location retrieval on mobile
- Implement backend endpoint/logic for receiving current location
- Store latest shared location with timestamp
- Handle denied permission cases
- Handle unavailable location cases
- Define how privacy settings affect location sharing

**Acceptance Criteria**
- Current device location can be requested and shared
- Backend stores shared location properly
- Permission denied and unavailable location cases are handled safely

**Dependencies**
- Shared map and location infrastructure

erinc00:
This PR completes the web scope of current-device-location sharing by making location retrieval permission-aware, sending structured location payloads to backend profile endpoints, and improving data integrity for stored location metadata.

## What’s Included
- Added permission-aware current location handling in the web location picker.
- Added safer geolocation error handling for:
  - permission denied
  - position unavailable
  - timeout
- Extended web location sharing payloads to include:
  - latitude / longitude
  - coordinate metadata (`source`, `accuracyMeters`, `capturedAt`)
  - administrative location fields
  - `displayAddress` and `placeId`
- Updated profile and complete-profile save flows to submit current location payload consistently.
- Preserved backend-provided location identity fields during profile hydration:
  - `placeId`
  - `displayAddress`
  - coordinate metadata continuity
- Expanded web profile API typings to match backend location response shape.
- Added e2e regression coverage to ensure saved location identity/metadata is not unintentionally overwritten.

## Files Changed
- LocationPicker.tsx
- ProfileView.tsx
- CompleteProfileForm.tsx
- profile.ts
- api.js
- profile-and-privacy.spec.js

## Notes
- Behavior is aligned with existing backend privacy/location separation:
  - location payload and privacy toggle remain separate update flows.

## Issue
Closes #290

### DIFF
--- a/backend/migrations/20260424_110000__add_location_coordinate_metadata_columns.sql
+++ b/backend/migrations/20260424_110000__add_location_coordinate_metadata_columns.sql
@@ -1,0 +1,18 @@
+BEGIN;
+
+ALTER TABLE location_profiles
+  ADD COLUMN IF NOT EXISTS coordinate_accuracy_meters DOUBLE PRECISION,
+  ADD COLUMN IF NOT EXISTS coordinate_source VARCHAR(100),
+  ADD COLUMN IF NOT EXISTS coordinate_captured_at TIMESTAMP;
+
+ALTER TABLE location_profiles
+  DROP CONSTRAINT IF EXISTS chk_location_profile_coordinate_accuracy;
+
+ALTER TABLE location_profiles
+  ADD CONSTRAINT chk_location_profile_coordinate_accuracy
+  CHECK (
+    coordinate_accuracy_meters IS NULL
+    OR coordinate_accuracy_meters >= 0
+  );
+
+COMMIT;
--- a/backend/src/modules/profiles/repository.js
+++ b/backend/src/modules/profiles/repository.js
@@ -72,6 +72,20 @@ function normalizeIsoCountryCode(value) {
   return /^[A-Z]{2}$/.test(normalized) ? normalized : null;
 }
 
+function normalizeOptionalTimestamp(value) {
+  const normalized = normalizeOptionalString(value);
+  if (!normalized) {
+    return null;
+  }
+
+  const parsedMs = Date.parse(normalized);
+  if (Number.isNaN(parsedMs)) {
+    return null;
+  }
+
+  return new Date(parsedMs).toISOString();
+}
+
 function normalizeLocationInput(data) {
   const administrative = isPlainObject(data.administrative) ? data.administrative : null;
   const coordinate = isPlainObject(data.coordinate) ? data.coordinate : null;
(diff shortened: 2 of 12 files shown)


kckagancan (changes requested):
Thanks for the changes @erinc00 . I found a few important issues that should be addressed before merge.

### Main problems

1. **“Share Current Location” does not actually guarantee a fresh device-location share**  
   The real device-location retrieval only happens when the user clicks **“Use Current Location”** in the location picker. The **“Share Current Location”** toggle only updates the privacy flag.  
   So a user can enable sharing and save the profile without ever refreshing their actual current device location.

2. **`capturedAt` is not the real geolocation capture time**  
   The saved timestamp is generated later in the frontend flow with `new Date().toISOString()`, instead of representing the actual time the device location was captured.  
   This makes the metadata less trustworthy.

3. **The PR improves location identity preservation, but full coordinate metadata continuity is not really proven**  
   Fields like `placeId` and `displayAddress` are better preserved now, which is good.  
   But the current end-to-end flow still does not clearly prove preservation of metadata like `source`, `accuracyMeters`, and a real capture timestamp.

4. **Test coverage does not validate the real risky cases of this feature**  
   The new tests mostly verify that saved location identity is not overwritten during profile save.  
   They do **not** really test the full current-device sharing flow, especially:
   - permission denied
   - position unavailable
   - timeout
   - enabling sharing without clicking “Use Current Location”
   - whether fresh device location metadata is actually persisted

5. **Minor metadata handling issue**  
   Some optional metadata hydration uses `||` instead of `??`, which can incorrectly drop valid falsy values such as `0`.

### Expected changes

- Make the share-current-location flow actually require or clearly coordinate with fresh device-location capture.
- Preserve the real geolocation capture timestamp instead of generating a later UI timestamp.
- Either reduce the metadata continuity claim or fully support and verify it end-to-end.
- Add tests for denied/unavailable/timeout cases and for the real current-device sharing behavior.
- Use `??` consistently for optional metadata fields.

kckagancan (changes requested):
Thanks for the follow-up changes @erinc00  — this is much closer now, and several of the previous issues were addressed well.
That said, I still see two important issues before merge.

### 1. Same-session second save on the profile page can still be blocked incorrectly
`ProfileView` now blocks first-time share enable unless there is a fresh `current_device` selection, which is good.

However, after a successful save:
- the page refreshes profile/location data from backend
- `locationPickerValue` is rehydrated from backend response
- but `initialShareLocation` is not updated after that refresh

At the same time, the current backend profile response still does **not** round-trip the coordinate metadata in a way that preserves `source: "current_device"` and the original fresh capture semantics.

So the flow can become:

1. user clicks **Use Current Location**
2. enables **Share Current Location**
3. saves successfully
4. page refreshes from backend
5. user edits something else and saves again in the same session
6. the guard can incorrectly block the save again

because the frontend no longer sees a fresh `current_device` selection, while `initialShareLocation` also still reflects the old pre-save state.

This looks like a real behavior bug.

### 2. I cannot confidently treat the new metadata-persistence E2E as proven yet
The new tests now expect the saved profile response to preserve things like:
- `coordinate.source === "current_device"`
- `accuracyMeters`
- a `capturedAt` value close to the real capture timestamp

That is the right direction.

However, based on the current backend profile mapping, the response still appears to map coordinate metadata in a way that does **not** obviously return those values end-to-end.

So before merge, I would want this to be explicitly verified with a green run on the latest head commit and not assumed from the frontend changes alone.

### What I’d want before approval
- Fix the profile-page same-session resave case so that successful first-time enable does not get blocked again afterward
- Make sure the latest head commit is actually green with the new E2E expectations
- If needed, update the frontend state refresh logic so it stays consistent with what the backend actually returns after save

kckagancan (approved):
Looks good now. Merging.

PR #336, opened 2026-04-23
feat: implement advanced help-request matching with multi-responder support

### LINKED ISSUE #264: [FEATURE] Implement advanced help-request matching algorithm`
The help-request matching logic should be improved to use multiple relevant signals such as urgency, location, skill, and availability where applicable. It should also behave more robustly in multi-volunteer, multi-emergency situations by reducing poor assignment outcomes.

**Tasks**
- Review current matching approach
- Define improved matching inputs
- Incorporate urgency into matching
- Incorporate location into matching
- Incorporate skill/availability where available
- Improve volunteer suitability evaluation
- Define deterministic and understandable matching behavior
- Improve multi-volunteer / multi-emergency allocation logic
- Reduce duplicate assignment behavior where possible
- Reduce neglected emergency outcomes where possible
- Consider current volunteer load in assignment decisions
- Improve fairness/load balancing across volunteers where feasible
- Test matching against representative scenarios
- Handle missing/partial data cases safely

**Acceptance Criteria**
- Matching uses intended signals such as urgency, location, skill, and availability
- Matching behavior is more robust and explainable than before
- Multi-volunteer and multi-emergency scenarios are handled more safely
- Duplicate or neglected assignment behavior is reduced compared to the previous state
- Missing/partial data does not break the flow

**Dependencies**
- Request help flow shape should be stable
- Relevant location-related data should exist first
- Assignment/deployment data readiness where applicable

kckagancan:
### Summary
This PR upgrades NEPH’s backend help-request matching from a simple single-responder flow into a more robust multi-signal, multi-responder matching system.

### What changed
- improved matching using help type, first-aid preference, proximity, and volunteer availability
- added deterministic first-pass assignment behavior
- added second-pass SAR expansion behavior
- added fallback handling for first-aid requests when no first-aid-capable volunteer is available
- preserved the existing public status model without introducing new public statuses
- fixed cancellation/rematching ordering issues
- enabled true multi-responder assignments per request via migration
- added active-assignment-per-volunteer protection
- updated request reads and resolve behavior to work more safely with multiple active assignments
- added/updated backend unit and integration coverage for the new matching behavior
- added migration safety coverage for environments that may already contain duplicate active assignments per volunteer

### Important behavior notes
- request assignment cardinality now supports multiple active responders per request
- first-aid-capable volunteers are preferred when a request still needs a first-aid specialist
- first-aid-only fallback growth is capped to avoid unnecessary general-volunteer over-assignment
- volunteer-side resolve now applies to that volunteer’s assignment, not automatically to the whole request
- `tryToAssignRequest(requestId)` is now request-scoped and no longer triggers a global matching cycle

### Compatibility / rollout notes
- frontend/mobile changes are not included in this PR
- no new public statuses were introduced
- requester-facing request reads still expose a single helper for compatibility, even though the backend now supports multiple active responders; richer multi-responder representation is follow-up work
- schema change is handled by migrations

Closes #264

### DIFF
--- a/backend/migrations/20260423_150000__guard_active_assignment_per_volunteer.sql
+++ b/backend/migrations/20260423_150000__guard_active_assignment_per_volunteer.sql
@@ -1,0 +1,7 @@
+BEGIN;
+
+CREATE UNIQUE INDEX IF NOT EXISTS idx_assignments_active_volunteer_unique
+  ON assignments(volunteer_id)
+  WHERE is_cancelled = FALSE;
+
+COMMIT;
--- a/backend/src/modules/availability/service.js
+++ b/backend/src/modules/availability/service.js
@@ -3,15 +3,43 @@ const {
   createVolunteer,
   updateVolunteerAvailability,
   createAvailabilityRecord,
+  findAvailableVolunteersForMatching,
   findMatchingRequestForVolunteer,
   createAssignment,
   updateRequestStatus,
   getAssignmentByVolunteerId,
   getAssignmentById,
-  findAssignmentByRequestId,
+  findActiveAssignmentsByRequestId,
   cancelAssignment,
-  findMatchingVolunteerForRequest,
 } = require('./repository');
+
+async function runAssignmentCycle() {
+  const availableVolunteers = await findAvailableVolunteersForMatching();
+  const createdAssignments = [];
+
+  for (const volunteer of availableVolunteers) {
+    const matchingRequest = await findMatchingRequestForVolunteer(volunteer.volunteer_id);
+
+    if (!matchingRequest) {
+      continue;
+    }
+
+    const assignment = await createAssignment(volunteer.volunteer_id, matchingRequest.request_id);
+    await updateRequestStatus(matchingRequest.request_id, 'ASSIGNED');
+    createdAssignments.push(assignment);
+  }
+
+  return createdAssignments;
+}
+
+async function syncRequestStatusFromAssignments(requestId) {
+  const activeAssignments = await findActiveAssignmentsByRequestId(requestId);
+  const nextStatus = activeAssignments.length > 0 ? 'ASSIGNED' : 'PENDING';
+
+  await updateRequestStatus(requestId, nextStatus);
+
+  return activeAssignments;
+}
 
 async function setAvailability(userId, { isAvailable, latitude, longitude }) {
   let volunteer = await findVolunteerByUserId(userId);
@@ -34,13 +62,8 @@ async function setAvailability(userId, { isAvailable, latitude, longitude }) {
   if (isAvailable) {
     const existingAssignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
     if (!existingAssignment) {
-      const matchingRequest = await findMatchingRequestForVolunteer(volunteer.volunteer_id);
-      if (matchingRequest) {
-        assignment = await createAssignment(volunteer.volunteer_id, matchingRequest.request_id);
-        await updateRequestStatus(matchingRequest.request_id, 'ASSIGNED');
-        // Refresh assignment with full data
-        assignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
-      }
+      await runAssignmentCycle();
+      assignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
     } else {
       assignment = existingAssignment;
     }
@@ -49,7 +72,8 @@ async function setAvailability(userId, { isAvailable, latitude, longitude }) {
     const activeAssignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
     if (activeAssignment) {
       await cancelAssignment(activeAssignment.assignment_id);
-      await updateRequestStatus(activeAssignment.request_id, 'PENDING');
+      await syncRequestStatusFromAssignments(activeAssignment.request_id);
+      await runAssignmentCycle();
     }
   }
 
@@ -88,16 +112,13 @@ async function syncAvailability(userId, { records }) {
 
   // If volunteer is now available and has no assignment, try to match
   if (updatedVolunteer.is_available && !currentAssignment) {
-    const matchingRequest = await findMatchingRequestForVolunteer(volunteer.volunteer_id);
-    if (matchingRequest) {
-      await createAssignment(volunteer.volunteer_id, matchingRequest.request_id);
-      await updateRequestStatus(matchingRequest.request_id, 'ASSIGNED');
-      currentAssignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
-    }
+    await runAssignmentCycle();
+    currentAssignment = await getAssignmentByVolunteerId(volunteer.volunteer_id);
   } else if (!updatedVolunteer.is_available && currentAssignment) {
     // If volunteer is now unavailable, cancel their active assignment
     await cancelAssignment(currentAssignment.assignment_id);
-    await updateRequestStatus(currentAssignment.request_id, 'PENDING');
+    await syncRequestStatusFromAssignments(currentAssignment.request_id);
+    await runAssignmentCycle();
     currentAssignment = null;
   }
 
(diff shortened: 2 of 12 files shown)


kckagancan:
Non-blocking follow-up: multi-responder assignment is now supported, but requester-facing help-request reads still expose only a single `helper` (the earliest active assignment). As a result, **My Help Requests** may show only the first assigned volunteer even when multiple responders are assigned. Worth tracking separately so the response/UI can represent multi-responder requests more accurately.

I will open an issue about this.

mehmetcangurbuz08 (changes requested):
This is a very strong PR overall. The matching redesign is thoughtful, test coverage is excellent, and the implementation is clearly heading in the right direction. Great job on deterministic behavior, multi-responder support, and preserving API/status compatibility.

That said, I’m marking this as Request Changes due to a few must-fix risk points before merge:

Preserve IN_PROGRESS status in reassignment/cancellation flows (must fix)
Current status sync logic can downgrade an IN_PROGRESS request back to ASSIGNED or PENDING when assignments change.
This is a behavioral regression risk and should be prevented by preserving IN_PROGRESS unless there is an explicit transition.

Add concurrency safety for assignment creation (must fix)
runAssignmentCycle performs multi-step matching/assignment without transactional locking or DB-level guardrails for “active assignment” uniqueness patterns.
Under concurrent requests, this can create duplicate/invalid active assignments.
Please add a robust DB-level or transaction-based protection strategy.

Cap fallback growth for first-aid-only requests (must fix)
In fallback scenarios (no first-aid-capable volunteer available), the request can keep attracting additional general volunteers before a specialist arrives.
This can lead to over-assignment and resource skew.
Please enforce a clear cap/rule (for example, allow only one general fallback responder for first-aid-only requests).

erinc00 (changes requested):
Thanks for the update. I reviewed the new matching logic again and found a few additional issues that I think should be addressed before merge.

## Findings

### 1. High: Any one responder can now resolve a multi-responder request and silently cancel every other active responder on it
**File:** `backend/src/modules/availability/service.js`

In `resolveMyAssignment`, the flow now marks the request as `RESOLVED` and then cancels **all** active assignments for that request.

That was acceptable in the old single-assignment model, but with the new multi-responder model it becomes a real behavioral regression: one volunteer can effectively end the whole response and force-drop the rest of the assigned team, even if other responders are still actively working.

This needs a clearer rule. Right now, “my resolve action” behaves more like “resolve for everyone.”


### 2. Medium: First-aid fallback requests can be starved of a specialist indefinitely under sustained load
**File:** `backend/src/modules/availability/repository.js`

From the current matching flow, partially covered requests are pushed into second-pass via `assignment_phase_rank`, while the first pass always exhausts `needs_initial_coverage` requests first.

That means if a first-aid request currently has only a general fallback responder, and there is a steady stream of fresh uncovered requests, a newly available first-aid-capable volunteer may keep getting assigned elsewhere instead of backfilling that fallback case.

This weakens the intended “fallback until specialist available” behavior, because in practice the specialist may never arrive under sustained load.


### 3. Medium: `tryToAssignRequest(requestId)` is no longer request-scoped and now mutates the whole matching system
**File:** `backend/src/modules/availability/service.js`

`tryToAssignRequest(requestId)` now runs the global `runAssignmentCycle()` and only then checks whether the target request ended up with any active assignment.

That means a caller trying to reassign one specific request can end up creating assignments for unrelated requests first. So despite the method name and old behavior suggesting a request-local operation, it now has system-wide side effects.

This makes targeted flows harder to reason about, harder to test, and more surprising for callers.

If this behavior is intentional, I think the method name/contract should reflect that. Otherwise it should be narrowed back to request-scoped behavior.

kckagancan:
Fixed both of the review points. Can you guys re-check please? @erinc00 @mehmetcangurbuz08

erinc00 (changes requested):
This is a substantial change and it is much more than a matching-quality improvement. The PR changes assignment cardinality, matching behavior, cancellation/rematching flow, and the operational meaning of an “assigned” request. There are good improvements here, but based on the visible diff I do not think this is low-risk yet.

Blocking
1.	resolveMyAssignment still appears to let a single volunteer resolve the whole request in a multi-responder model
backend/src/modules/availability/service.js
The new flow looks better than the earlier version because it no longer cancels every active assignment on the request. However, it still sets the request status to RESOLVED when a single volunteer resolves their own assignment, then cancels only that volunteer’s assignment and re-syncs status afterward.
In the old single-assignment model this was acceptable. In the new multi-responder model it is no longer safe by default. A volunteer completing their own part is not necessarily the same thing as the whole request being resolved. If multiple responders are active on the same request, request-level resolution should not be silently triggered by one responder unless that is an explicit product decision.
Right now the behavior looks semantically wrong for multi-responder coordination and can produce inconsistent lifecycle transitions.

2.	The backend now supports multiple active responders per request, but public reads still expose only a single helper
backend/src/modules/help-requests/repository.js
The read path now uses a lateral subquery to pick only the earliest active assignment as the “primary” assignment. That avoids row explosion, but it also means the backend can hold multiple active responders while clients still receive a single-helper view of that request.
So after this PR, the system can be in a true multi-responder state while clients still receive a single-helper view. That is not just “preserving API shape”; it is a behavioral contract drift. The visible state and the actual state can diverge.
If multi-responder is truly intended, the read model should either expose that explicitly or the domain decision should be reconsidered. As it stands, requesters and other clients may be shown an incomplete picture of who is actually assigned.

3.	The new active-assignment-per-volunteer unique index can fail on existing data
backend/migrations/20260423_150000__guard_active_assignment_per_volunteer.sql
Adding a partial unique index on active assignments per volunteer is the right direction. However, there is no visible cleanup or validation step before creating it. If existing environments already contain more than one active assignment for the same volunteer, this migration can fail during deployment.
Given the scope of this PR, that risk should be handled explicitly rather than assumed away.

High

1.	The PR description understates the actual behavioral change
The summary says the current public API shape and status behavior are preserved, and that no new public statuses were introduced. Technically the response shape may remain similar, but the system behavior absolutely changes in meaningful ways:
	•	request-level assignment cardinality changes
	•	resolution semantics change
	•	reassignment behavior changes
	•	request reads now hide multi-responder reality behind a single “primary” helper
This is not a small internal optimization. It is a domain-level behavior change, and it should be treated that way in review and rollout.

2.	tryToAssignRequest(requestId) now behaves as a multi-fill operation, not a single targeted assignment
backend/src/modules/availability/service.js
The new implementation loops until there are no more eligible volunteers for the target request. That means this function is no longer “assign one volunteer if possible”; it now means “fill this request with all eligible responders according to current repository logic.”
That may be intended, but it is a major semantic change under the same method name. If some request types are still expected to remain single-responder, that responsibility is now entirely pushed into repository selection logic. That increases the risk surface significantly.

3.	Cancellation by request now triggers wider system-side effects than the method name suggests
backend/src/modules/availability/service.js
cancelAssignmentByRequestId now cancels all active assignments on the request and then runs the matching cycle again, which can immediately redistribute the freed volunteers to other requests. That may be operationally reasonable, but the method is no longer request-local in effect.
This makes debugging and reasoning harder because a call that sounds like scoped cleanup now also causes broader system mutation.

Medium

1.	Request status writes are still spread across too many paths
Files: backend/src/modules/availability/service.js, backend/src/modules/help-requests/service.js
There are now multiple overlapping mechanisms controlling request status:
	•	direct updateRequestStatus(..., 'ASSIGNED')
	•	direct updateRequestStatus(..., 'RESOLVED')
	•	markRequestAssignedIfPending(...)
	•	syncRequestStatusPreservingInProgress(...)
	•	request cancel flows
	•	request resolve flows
	•	matching cycle side effects
The current PR improves some ordering issues, but the status lifecycle is becoming more distributed and therefore easier to break later.

2.	The “primary helper” selection rule is deterministic but not clearly justified as business logic
backend/src/modules/help-requests/repository.js
If multiple responders are active, the public helper shown to clients is chosen as the earliest active assignment. That is deterministic, but it is not obviously the right business rule. It could just as easily have been nearest responder, first-aid-capable responder, coordinator-designated lead responder, or some other explicit policy.
The implementation chooses one policy, but the rationale is not visible here.

3.	There may be dead or outdated imports in the service layer after the refactor
backend/src/modules/availability/service.js
From the visible hunk, findAssignmentByRequestId still appears in the import list even though the new flow seems to rely on findActiveAssignmentsByRequestId. This is not important by itself, but in a PR this large it is a sign that some cleanup may still be incomplete.

erinc00 (changes requested):
Great work overall on this branch.

This PR is a strong improvement in matching quality and test coverage.  
However, I found a few important issues that should be addressed before merge.

## Findings

### 1. Potential blocker: resolved requests may leave active assignment rows, which can silently block future matches
In the current flow, requester/guest resolve updates request status but does not always clear active assignment rows.  
With the new unique partial index on active assignments per volunteer, a volunteer can remain “occupied” by a stale active row even after request resolution.

**Impact**
- volunteer may fail to receive new assignments
- insert may no-op due to ON CONFLICT, making this hard to detect in runtime behavior

### 2. Migration cleanup can keep the wrong assignment row
The reconciliation migration keeps one active assignment per volunteer using only assignment timestamp/id order.  
It does not prioritize assignment rows attached to still-open requests over stale/closed contexts.

**Impact**
- a stale row can be kept while a useful one is removed
- together with the active-assignment guard, this can lock matching state in bad historical datasets

### 3. Matching request query may duplicate candidates if request has multiple location rows
Open-request matching joins request locations without deterministic “latest location” selection.

**Impact**
- same request can appear multiple times in candidate pool
- ranking behavior may become data-shape dependent and less deterministic

erinc00 (approved):
Good effort on incorporating the feedback. The updates look solid and the issues have been resolved. Approving and merging.

PR #342, opened 2026-04-23
feat: add admin dashboard page with guarded access and emergency overview

### LINKED ISSUE #269: [FEATURE][Dashboard] Build aggregate emergency overview

mehmetcangurbuz08:
This PR adds the web-side admin dashboard for aggregate emergency visibility and connects it to the new backend admin overview endpoint.

What was implemented
Added a new admin page at /admin with dashboard cards for:
total / active / resolved / closed emergency counts
status breakdown
urgency breakdown
recent activity summary
optional region summary table
Added a dedicated AdminRouteGate to enforce admin-only access on the client.
Added an admin API client module to fetch overview data from:
/api/admin/emergency-overview
/api/admin/emergency-overview?includeRegionSummary=true
Integrated loading, error, and empty-state handling in the dashboard view.
Updated top navigation to show the Admin link only for authenticated admin users.
UX and behavior details
Non-authenticated users are redirected to login with returnTo=/admin.
Authenticated non-admin users are redirected to /home.
Region summary is fetched only when requested, matching backend optional behavior.
Refresh-loop behavior in the region summary area was fixed by separating initial load and refresh states.
Files added
web/src/app/admin/page.tsx
web/src/components/auth/AdminRouteGate.tsx
web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
web/src/lib/admin.ts
Files updated
web/src/components/layout/TopNavbar.tsx
web/src/styles/globals.css
Notes
This PR assumes admin users already exist in the database and can authenticate via the existing login flow.
The dashboard now reflects real backend aggregate data once emergency records are present.

Important setup note
To access the admin page, the logged-in user must exist in the admins table.
If needed, add the user manually in DB:

INSERT INTO admins (admin_id, user_id, role)
VALUES ('admin_manual_1', '<EXISTING_USER_ID>', 'COORDINATOR');

Closes #269

### DIFF
--- a/web/src/app/admin/page.tsx
+++ b/web/src/app/admin/page.tsx
@@ -1,0 +1,16 @@
+import { Suspense } from "react";
+import { AppShell } from "@/components/layout/AppShell";
+import { AdminRouteGate } from "@/components/auth/AdminRouteGate";
+import AdminEmergencyOverviewView from "@/components/feature/admin/AdminEmergencyOverviewView";
+
+export default function AdminPage() {
+    return (
+        <Suspense>
+            <AdminRouteGate>
+                <AppShell title="Admin Dashboard">
+                    <AdminEmergencyOverviewView />
+                </AppShell>
+            </AdminRouteGate>
+        </Suspense>
+    );
+}
--- a/web/src/app/admin/page.tsx
+++ b/web/src/app/admin/page.tsx
@@ -1,16 +1,13 @@
-import { Suspense } from "react";
 import { AppShell } from "@/components/layout/AppShell";
 import { AdminRouteGate } from "@/components/auth/AdminRouteGate";
 import AdminEmergencyOverviewView from "@/components/feature/admin/AdminEmergencyOverviewView";
 
 export default function AdminPage() {
     return (
-        <Suspense>
-            <AdminRouteGate>
-                <AppShell title="Admin Dashboard">
-                    <AdminEmergencyOverviewView />
-                </AppShell>
-            </AdminRouteGate>
-        </Suspense>
+        <AdminRouteGate>
+            <AppShell title="Admin Dashboard">
+                <AdminEmergencyOverviewView />
+            </AppShell>
+        </AdminRouteGate>
     );
 }
--- a/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
+++ b/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
@@ -287,7 +287,9 @@ export default function AdminEmergencyOverviewView() {
                 </div>
 
                 {includeRegionSummary ? (
-                    regionSummary.length > 0 ? (
+                    regionError ? (
+                        <p className="admin-error-text">Region summary refresh failed: {regionError}</p>
+                    ) : regionSummary.length > 0 ? (
                         <div className="admin-region-table-wrap">
                             <table className="admin-region-table">
                                 <thead>
@@ -324,10 +326,6 @@ export default function AdminEmergencyOverviewView() {
                         Region summary is optional. Load it when you need city-level details.
                     </p>
                 )}
-
-                {includeRegionSummary && regionError ? (
-                    <p className="admin-error-text">Region summary refresh failed: {regionError}</p>
-                ) : null}
             </SectionCard>
         </div>
     );
--- a/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
+++ b/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
@@ -50,6 +50,8 @@ export default function AdminEmergencyOverviewView() {
         async (includeRegion: boolean, mode: "initial" | "refresh" = "refresh") => {
             const token = getAccessToken();
             if (!token) {
+                setLoading(false);
+                setRefreshing(false);
                 redirectToLogin();
                 return;
             }
@@ -115,6 +117,23 @@ export default function AdminEmergencyOverviewView() {
         },
         [redirectToLogin, router]
     );
+
+    React.useEffect(() => {
+        if (!loading) {
+            return;
+        }
+
+        const timeoutId = window.setTimeout(() => {
+            setLoading(false);
+            setInitialError((currentError) =>
+                currentError || "Request timed out. Please try again."
+            );
+        }, 16_000);
+
+        return () => {
+            window.clearTimeout(timeoutId);
+        };
+    }, [loading]);
 
     React.useEffect(() => {
         const mode = isFirstLoadRef.current ? "initial" : "refresh";
--- a/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
+++ b/web/src/components/feature/admin/AdminEmergencyOverviewView.tsx
@@ -1,0 +1,252 @@
+"use client";
+
+import * as React from "react";
+import { usePathname, useRouter } from "next/navigation";
+import { ApiError } from "@/lib/api";
+import { getAccessToken, clearAccessToken } from "@/lib/auth";
+import { fetchAdminEmergencyOverview, type EmergencyOverview } from "@/lib/admin";
+import { SectionCard } from "@/components/ui/display/SectionCard";
+import { SectionHeader } from "@/components/ui/display/SectionHeader";
+import { PrimaryButton } from "@/components/ui/buttons/PrimaryButton";
+import { SecondaryButton } from "@/components/ui/buttons/SecondaryButton";
+
+function MetricTile({
+    label,
+    value,
+    tone = "default",
+}: {
+    label: string;
+    value: number;
+    tone?: "default" | "success" | "warning" | "danger";
+}) {
+    return (
+        <article className={`admin-metric-tile tone-${tone}`}>
+            <p className="admin-metric-label">{label}</p>
+            <p className="admin-metric-value">{value}</p>
+        </article>
+    );
+}
+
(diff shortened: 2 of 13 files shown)


erinc00 (changes requested):
Great progress overall — the page structure is clear, the backend contract is used cleanly, and the optional region-summary flow is reflected in the UI reasonably well. That said, after reviewing the implementation and the current branch state, I do not think this is ready to merge yet.

## Blocking / Must Fix


### 1. There is no actual web test coverage for the new admin flow

The PR adds:

* a new `/admin` page
* a new `AdminRouteGate`
* new navbar admin-link logic
* optional region-summary fetching behavior

But from what is shown, there are **no new web E2E/component tests covering this feature**.

That is a major gap, especially because this PR changes access control and route behavior on the client side.

At minimum I would expect coverage for:

* guest visiting `/admin` → redirected to login with `returnTo=/admin`
* authenticated non-admin visiting `/admin` → redirected to `/home`
* authenticated admin visiting `/admin` → dashboard loads
* admin link visibility in navbar:

  * hidden for guest
  * hidden for authenticated non-admin
  * shown for authenticated admin
* region summary toggle:

  * default state does not render table
  * enabling it triggers the region-summary fetch path
  * disabling it hides region-summary UI again
* retry flow on initial overview fetch error

Without those, the core behavior introduced by this PR is mostly untested.

---

## High Priority Findings

### 1. `AdminRouteGate` and `TopNavbar` duplicate admin/auth resolution separately

**Files:**

* `web/src/components/auth/AdminRouteGate.tsx`
* `web/src/components/layout/TopNavbar.tsx`

Both components independently:

* read the token
* call `fetchCurrentUser`
* decide whether the user is admin
* clear the token on failure
* redirect or update UI

This creates two problems:

#### a. duplicated network work

Opening `/admin` can cause multiple `/me`-style checks in parallel or close succession.

#### b. inconsistent behavior risk

If one request fails transiently while another succeeds, you can get odd UI transitions:

* navbar clears auth while route gate is still resolving
* gate redirects while navbar still thinks user is logged in
* transient `/me` failure logs the user out even though the token may still be valid

This may not break all the time, but it is fragile.

**Suggestion:**
Use a shared auth/admin state source (context/store/session hook) rather than having both navbar and route gate re-implement session resolution independently.

---

### 2. `TopNavbar` can now log users out on transient `/me` failures

**File:** `web/src/components/layout/TopNavbar.tsx`

The navbar now calls `fetchCurrentUser(token)` during:

* initial mount
* `storage`
* `focus`
* custom `neph-auth-changed`
* `visibilitychange`

and in the `catch` branch it calls:

```ts
clearAccessToken();
setIsAuthenticated(false);
setIsAdmin(false);
```

That means a temporary backend issue, network hiccup, or short-lived /me failure can cause the navbar to log the user out client-side.

That is a pretty aggressive response for what may just be a transient fetch failure.

Suggestion:
Differentiate between:
•	true auth failure (401/invalid token)
•	transient request failure (network / 5xx / temporary fetch issue)

Only clear the token for actual auth-invalid cases.

⸻

Medium Priority Findings

1. AdminEmergencyOverviewView has no request ordering protection

web/src/components/feature/admin/AdminEmergencyOverviewView.tsx

loadOverview is called from an effect whenever includeRegionSummary changes. There is no cancellation or request versioning.

So if the user toggles the region-summary flag quickly:
•	request A starts
•	request B starts
•	response A returns after B
•	stale response A can overwrite newer state

This is a classic async race condition.

It may not happen often, but it is possible and easy to avoid.

Suggestion:
Track request identity or cancellation so only the latest request updates state.

⸻

2. Error handling is not scoped cleanly once overview data already exists

web/src/components/feature/admin/AdminEmergencyOverviewView.tsx

If the dashboard has already loaded successfully and a later refresh fails:
•	the previous overview stays rendered
•	a generic error card appears at the bottom

That is not inherently wrong, but the UX is a bit ambiguous:
•	are the visible numbers fresh or stale?
•	did the main overview fail or only the region-summary refresh fail?
•	should the stale data remain visible without any “showing previous data” indication?

This is especially relevant for the optional region-summary toggle/refresh path.

Suggestion:
Distinguish between:
•	initial load failure
•	refresh failure while showing previous data
•	region-summary-specific failure

Even a small “Showing previous data; refresh failed” hint would make the state clearer.

⸻

3. The PR says “empty-state handling” exists, but the view does not really model a distinct successful-empty state

web/src/components/feature/admin/AdminEmergencyOverviewView.tsx

Right now the view has:
•	loading
•	no overview → error-ish empty fallback
•	overview present → render dashboard

But for an aggregate dashboard, a “successful but empty system state” is different from “failed to load overview data”.

Example:
•	no emergencies in the system
•	no recent activity
•	no region rows

That is still a valid successful state, not an empty/error fallback.

Currently this is handled implicitly by rendering zeroes, which may be acceptable, but then the PR description should not overstate “empty-state handling” unless that state is intentionally modeled.

Suggestion:
Either:
•	keep the current behavior and simplify the PR wording, or
•	explicitly define/render a meaningful zero-data state

⸻

4. Region summary refresh is coupled to the full overview fetch

web/src/components/feature/admin/AdminEmergencyOverviewView.tsx

Toggling region summary re-fetches the full overview from:
•	/api/admin/emergency-overview
•	/api/admin/emergency-overview?includeRegionSummary=true

This is consistent with the backend API, but from a UI perspective it means the “Load Region Summary” action is not actually a narrow region-only request; it is a full overview refresh.

That may be okay for now, but it is worth acknowledging because:
•	it increases unnecessary data refetching
•	it makes “refreshing” messaging ambiguous
•	it broadens the effect of a region-summary toggle

Not a blocker, but something to keep in mind.

⸻

5. Suspense is used without a fallback and does not seem to add value here

web/src/app/admin/page.tsx

```tsx
<Suspense>
    <AdminRouteGate>
        <AppShell title="Admin Dashboard">
            <AdminEmergencyOverviewView />
        </AppShell>
    </AdminRouteGate>
</Suspense>
```

There is no fallback provided, and nothing shown here obviously requires this wrapper in its current form.

Not a functional bug, but it looks unnecessary unless there is a specific reason tied to router/search-param behavior.

⸻

Low Priority / Polish

1. Region summary table uses raw city strings directly

web/src/components/feature/admin/AdminEmergencyOverviewView.tsx

The table renders item.city directly. If backend normalization is inconsistent or historical data varies, the UI may end up showing:
•	ankara
•	Ankara
•	ANKARA

This is more of a backend/data concern, but the dashboard will surface it visibly.

⸻

2. Navbar admin-link visibility depends on async user fetch after auth detection

web/src/components/layout/TopNavbar.tsx

A logged-in admin user will not immediately see the Admin link until fetchCurrentUser resolves. That is probably acceptable, but it does mean the nav can temporarily render as if the user is non-admin.

Not blocking, just worth noting.

mehmetcangurbuz08:
@copilot resolve the merge conflicts in this pull request

kckagancan (changes requested):
Thanks for the implementation @mehmetcangurbuz08 . I still found a few important issues before merge.

### Main problems

1. **`AdminRouteGate` is not actually checking admin role**
   The gate only checks whether an access token exists. It does not verify whether the authenticated user is an admin.
   
   So non-admin authenticated users can still enter the admin page shell until the overview request returns `403` and redirects them to `/home`.

   Backend protection exists, so this is not a backend security hole, but the client-side gate is incomplete and the PR description is stronger than the actual implementation.

2. **Region summary mixes error state with empty state**
   When region summary is enabled and that request fails, the UI can still show:
   - `No regional records available.`
   - and also `Region summary refresh failed: ...`

   A failed request should not be shown as if it were a valid empty result. This is a real state-handling bug.

3. **Aggregate metric correctness is not tested deeply enough**
   The E2E coverage is good for:
   - redirects
   - navbar visibility
   - retry behavior
   - region summary toggling

   But it does not strongly verify that the actual aggregate numbers are correct across different combinations of:
   - statuses
   - urgency levels
   - recent activity windows
   - closed-count behavior

4. **The “Checking admin access...” copy is misleading**
   The loading message in `AdminRouteGate` says it is checking admin access, but it is only checking token presence.

5. **Loading/retry UX is a bit mixed**
   The initial loading state also immediately shows a retry affordance. Not a major bug, but the UX is a little inconsistent.

### Expected changes

- Make `AdminRouteGate` actually verify admin role on the client.
- Fix region summary so failed fetches are shown only as error state, not as empty state too.
- Add stronger validation for aggregate metric correctness.
- Adjust the access-check/loading copy so it matches the real behavior.

erinc00 (approved):
Reviewed the PR in detail. Code is clean, logic is correct, and no issues found.  
Approving and merging

PR #350, opened 2026-04-26
feat(admin): add deployment monitoring signals and dashboard tab

erinc00:
This PR adds an end-to-end Deployment Monitoring feature for admin users.

It introduces a new admin API and a new Admin dashboard tab to surface problematic emergency assignment situations with configurable thresholds.

## What’s Added
- New backend endpoint: `GET /admin/deployment-monitoring`
- New monitoring signals:
  - Unassigned emergencies
  - Long-waiting emergencies
  - In-progress emergencies
  - Potentially neglected emergencies
  - Conflicting / duplicate active reports
- Threshold/query controls:
  - `waitThresholdHours`
  - `neglectThresholdHours`
  - `listLimit`
- New web admin view with:
  - Monitoring controls
  - Signal summary tiles
  - Per-signal tables
  - Conflict grouping cards
- Admin dashboard switcher integration with new **Deployment Monitoring** tab
- Test coverage:
  - Backend integration tests for deployment monitoring
  - E2E Playwright spec for admin monitoring tab + retry/recovery flow

## Important Design Note
Conflict detection is implemented as **duplicate active reports** (grouped by city + need type + contact within recent window), instead of volunteer-overload, because the schema already enforces one active assignment per volunteer via partial unique index.

### DIFF
--- a/backend/src/modules/admin/controller.js
+++ b/backend/src/modules/admin/controller.js
@@ -6,6 +6,7 @@ const {
   getEmergencyOverviewForAdmin,
   getEmergencyHistoryForAdmin,
   getEmergencyAnalyticsForAdmin,
+  getDeploymentMonitoringForAdmin,
 } = require('./service');
 
 const ALLOWED_HISTORY_STATUSES = new Set(['RESOLVED', 'CANCELLED']);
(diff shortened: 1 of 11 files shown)


kckagancan (changes requested):
Thanks for the work on this PR @erinc00 .I found one behavioral issue that should be fixed before approval.

### 1. Duplicate/conflict groups can be incorrectly merged in the response

The SQL detects duplicate active reports by grouping on:

- city
- need type
- contact phone

That matches the PR description and makes sense for duplicate report detection.

However, after fetching the conflict rows, the JS response grouping uses only:

    city + needType

It does not include `contact_phone` in the response group key.

This means two separate duplicate groups in the same city with the same need type but different contact phones can be merged into one conflict group. In that case:

- unrelated duplicate groups are displayed together
- `duplicateCount` can become inaccurate
- the summary conflict count can disagree with the visible grouped cards
- admins may see misleading conflict information

Example scenario:

- Istanbul + first_aid + phone A has 2 duplicate reports
- Istanbul + first_aid + phone B has 2 duplicate reports

The backend summary would count 2 conflict groups, but the returned grouped list may collapse them into one Istanbul / first_aid group.

Please include the contact key in the conflict grouping identity, or return a safe masked/contact-derived group identifier.

For example, the backend grouping key should effectively be based on:

    city + needType + contactPhone

or a non-sensitive equivalent.


## E2E Failure

The failure is in an existing profile/privacy test:

    e2e/profile-and-privacy.spec.js
    profile save blocks first-time share enable until Use Current Location is used

The failing assertion is:

    Expected Share Current Location aria-pressed="true"
    Received aria-pressed="false"

This does not look directly related to the deployment monitoring code because this PR does not modify profile/privacy code. It may be a flaky or existing failing test, but CI is still red and should not be ignored before merge.

kckagancan (approved):

PR #351, opened 2026-04-26
fix(auth): scope login rate-limit per email+ip

### LINKED ISSUE #189: [BUG] Too many request error while log in
### Bug Description

 Check for this bug :Too many request error while log in

### Steps to Reproduce

1. Verificate Email
2. Press log in button too many times

### Expected Behavior

It gives too many request error

### Actual Behavior

Not reload after 1 minute

### Severity

Medium

delibas-rojhat:
Fixes #189

## What changed
- Scoped login rate limit key to `ip + normalized email` instead of global IP-only blocking.
- Added a dedicated login limiter with:
  - `windowMs: 1 minute`
  - `max: 5`
  - `skipSuccessfulRequests: true`
- Kept existing auth limiter behavior for other auth endpoints.
- Made key generation IPv6-safe with express-rate-limit helper.

## Why
Previously, after too many failed attempts, users could remain blocked in a way that affected subsequent login attempts broadly.  
Now lockout is isolated per email+IP, so one account’s failures don’t effectively freeze all login attempts from that client.

## Tests
- Added unit test: `backend/tests/unit/modules/auth/routes.rate-limit.test.js`
  - verifies key format behavior
  - verifies 6th failed attempt is blocked for same email
  - verifies different email is not blocked
- Auth unit tests passed locally.

### DIFF
--- a/backend/src/modules/auth/routes.js
+++ b/backend/src/modules/auth/routes.js
@@ -3,13 +3,41 @@ const rateLimit = require('express-rate-limit');
 const { env } = require('../../config/env');
 const { adminRouter } = require('../admin/routes');
 
+function shouldSkipAuthRateLimit() {
+  return env.nodeEnv === 'test' && process.env.ENABLE_RATE_LIMIT_IN_TEST !== 'true';
+}
+
+function buildLoginRateLimitKey(request) {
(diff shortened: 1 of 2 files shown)


kckagancan (approved):
Thanks for the fix. Looks fine. Merging.

PR #354, opened 2026-04-26
feat: improve requester request overview across current and history

### LINKED ISSUE #339: [FEATURE] Add aggregated emergency/request overview views instead of only per-request views
### Description
Current request/emergency flows are mostly centered around individual request views. Based on the MVP feedback, the system should also support **aggregated views** where multiple emergencies/requests can be seen together in a meaningful way instead of only one by one.

This is needed so users/admins/volunteers can understand the general emergency landscape at a glance.

### Why this is needed
Feedback indicated that data should not live only in single emergency/request pages. There should also be **aggregated views** that allow multiple emergencies to be reviewed together.

This becomes even more important after the advanced matching work, because the system now has richer assignment behavior and more request state complexity.

### Scope
This issue focuses on creating or improving pages/components that show multiple requests together.

### Possible Affected areas:
- My Help Requests
- Assigned Request / Assigned Requests
- list/table/card views that summarize multiple requests at once

### Tasks
- identify where the product currently only supports single-request understanding
- design aggregated request/emergency views for relevant user roles
- define which fields should be visible in list/overview mode
- update backend/frontend data flow if the current response shape is not enough
- ensure aggregated views remain readable and useful under multiple simultaneous requests

### Acceptance Criteria
- users can view multiple emergencies/requests together in an aggregated view
- the aggregated view is not just a raw list, but includes meaningful summary information
- request/emergency overview becomes easier to understand at a glance
- relevant screens are aligned with the newer advanced matching model

### Notes
- This issue is about **aggregate visibility**, not full request history and not detailed lifecycle metadata by itself
- Detailed status/priority/aging visibility is covered separately
- This is a sub-issue under the parent issue for improving request-related pages and backend response flows after advanced matching

kckagancan:
### Summary
This PR improves requester-side request overview in the Android `My Help Requests` flow by making the screen more informative at a glance across the current request and request history.

### Why this PR exists
Issue #339 is about aggregate request visibility in request-related flows, not admin analytics and not detailed lifecycle/history work by itself.

After reviewing the current implementation, the main requester-side foundations were already present:
- current vs history grouping already existed
- request ordering and metadata were already available
- past requests were already preserved locally

The remaining gap was that the screen still leaned too much toward isolated per-request reading and lacked a lightweight overview layer above the grouped sections.

### What changed
- added a small requester overview builder in `MyHelpRequestsRepository`
- added a lightweight overview card in `MyHelpRequestsScreen`
- improved aggregate visibility across the current request and request history
- preserved the existing current/history structure and styling
- added regression coverage for requester overview behavior

### Important behavior notes
- this PR focuses on requester-side aggregate visibility
- it does **not** introduce support for multiple simultaneous active requests
- it stays compatible with the current business rule that a requester has at most one active request at a time
- the overview value comes from making the current + historical request landscape easier to understand together
- no admin dashboard work was added
- no backend changes were required
- no advanced matching logic was changed

### Scope notes
- no schema changes
- no migration changes
- no history redesign
- no broad lifecycle/priority redesign
- no unrelated UI refactor
- minimal Android-only improvement to the existing requester overview surface

Closes #339

### DIFF
--- a/android/app/src/main/java/com/neph/features/myhelprequests/data/MyHelpRequestsRepository.kt
+++ b/android/app/src/main/java/com/neph/features/myhelprequests/data/MyHelpRequestsRepository.kt
@@ -59,6 +59,24 @@ data class MyHelpRequestUiModel(
         get() = syncStatus == SyncStatus.FAILED || syncStatus == SyncStatus.CONFLICTED
 }
 
+data class MyHelpRequestsOverviewUiModel(
+    val totalRequests: Int,
+    val activeRequests: List<MyHelpRequestUiModel>,
+    val historyRequests: List<MyHelpRequestUiModel>,
+    val resolvedCount: Int,
+    val cancelledCount: Int,
+    val assignedResponderCount: Int
+) {
+    val activeCount: Int
+        get() = activeRequests.size
+
+    val historyCount: Int
+        get() = historyRequests.size
+
+    val hasMultipleRequestContext: Boolean
+        get() = totalRequests > 1 || historyCount > 0
+}
+
 data class AssignedResponderUiModel(
     val firstName: String?,
     val lastName: String?,
@@ -183,6 +201,22 @@ object MyHelpRequestsRepository {
             now = System.currentTimeMillis()
         ).toUiModel()
     }
+}
+
+internal fun buildMyHelpRequestsOverview(
+    requests: List<MyHelpRequestUiModel>
+): MyHelpRequestsOverviewUiModel {
+    val activeRequests = requests.filter { it.isActive }
+    val historyRequests = requests.filterNot { it.isActive }
+
+    return MyHelpRequestsOverviewUiModel(
+        totalRequests = requests.size,
+        activeRequests = activeRequests,
+        historyRequests = historyRequests,
+        resolvedCount = requests.count { it.status.trim().uppercase() == "RESOLVED" },
+        cancelledCount = requests.count { it.status.trim().uppercase() == "CANCELLED" },
+        assignedResponderCount = activeRequests.sumOf { request -> request.responders.size }
+    )
 }
 
 internal fun HelpRequestEntity.toUiModel(): MyHelpRequestUiModel {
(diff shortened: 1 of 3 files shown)


delibas-rojhat (changes requested):
Thanks for the work on this PR @kckagancan . The requester overview improvements are valuable. I still see a few points that should be addressed before merge:

### 1) Shared resolve/action state can leak across multiple active request entries (Medium)

The screen now renders multiple active request entries, but resolve/loading/message state is still stored in single shared variables (`actionInProgressRequestId`, `actionMessageRequestId`, `actionMessage`).

When users trigger actions quickly on different entries, feedback can appear on the wrong entry (loading/message mismatch).

Please scope these states per `requestId` (map-based or per-entry state) so each request entry has isolated action feedback.

### 2) Current behavior may conflict with a “single active request” invariant (Medium, product-rule dependent)

This PR presents multiple active requests as a supported steady state (plural section text, overview metrics, and tests).

If our product rule is still “at most one active request per requester,” this is a rule mismatch.

Please either:
- enforce/restore the single-active invariant, or
- explicitly confirm that multi-active is now intended behavior and align docs/tests accordingly.

### 3) MATCHED mapper test does not validate real mapper input (Low)

In `MyHelpRequestsRepositoryMappingTest`, MATCHED behavior is asserted from a `copy(...)`-derived model rather than a real `HelpRequestEntity(status="MATCHED")` mapped through `toUiModel()`.

This can miss mapper regressions in actual conversion logic.

Please add a direct entity-to-mapper assertion for MATCHED status.

kckagancan:
I fixed the problems you addressed. Can you re-review? @delibas-rojhat

delibas-rojhat (changes requested):
Thanks for the update. I re-checked this and there is still one merge-blocking issue.

### 1) Android compile failure from PR changes (Blocker)

This branch fails on Android compile with:

`MyHelpRequestsScreen.kt:15:43 Cannot access 'val RowColumnParentData?.weight: Float': it is internal in file.`

So this PR is currently not mergeable.

Please fix the `weight` usage/import in `MyHelpRequestsScreen.kt` and re-run checks. I can re-review right after.

delibas-rojhat (approved):
Thanks for the quick fix. I re-checked the latest changes and the previous Android compile blocker is resolved.

The update looks good to me now, approving.

PR #399, opened 2026-05-01
feat(admin): complete user ban and unban moderation flow with auth enforcement and assignment/request side effects

### LINKED ISSUE #397: [FEATURE] Admin User Moderation: Ban/Unban users
Description  
Implement admin moderation actions to ban and unban users from the Admin Dashboard Users tab, with optional ban reason input in a confirmation modal.

Problem  
There is no enforceable account-level moderation control to block abusive or problematic users.

Scope  
- Add ban/unban capability for admins.
- Store ban state in user data.
- Enforce ban at authentication/authorization level.
- Allow optional ban reason entry via modal during ban action.

Database  
Add fields to `users` table via SQL migration:
- `is_banned` boolean NOT NULL DEFAULT false
- `ban_reason` text NULL
- `banned_at` timestamp NULL

Important:
- Use migration file under migrations.
- Do not modify init.sql for routine schema evolution.

Backend  
Admin-only endpoints:
- `PATCH /api/admin/users/:userId/ban`
- Body: `{ "reason": "optional string" }`
- `PATCH /api/admin/users/:userId/unban`

Ban behavior:
- Set `is_banned = true`
- Set `ban_reason = provided reason or null`
- Set `banned_at = now` 

Unban behavior:
- Set `is_banned = false`
- Clear `ban_reason` and `banned_at`

Auth enforcement:
- Banned user cannot log in.
- If user is already logged in and then banned, protected routes should reject with 403 on subsequent requests.

Admin UI  
In Users tab:
- Show Ban button for non-banned users.
- Show Unban button for banned users.
- Ban action opens confirmation modal with:
- optional textarea: “Reason (optional)”
- actions: Cancel / Confirm Ban
- Display ban status badge and reason in table/detail view.

Acceptance Criteria  
1. Migration adds `is_banned` (and optional reason/timestamp fields).
2. Ban endpoint is admin-only and supports optional reason.
3. Unban endpoint is admin-only and restores access.
4. Banned users cannot log in.
5. Banned users with active token receive 403 on protected endpoints.
6. Admin UI supports Ban/Unban with correct conditional action buttons.
7. Ban modal supports optional reason and submits successfully with/without text.
8. UI reflects latest user state after action without stale data.
9. Backend tests cover:
- admin authorization
- ban/unban state transitions
- login/access rejection for banned users
10. Web E2E covers:
- admin banning a user
- banned user blocked from login/access
- admin unbanning user
- user regains login/access

erinc00:
This PR delivers the full scope of the admin user ban and unban moderation feature for this branch, including backend APIs, auth enforcement, admin UI actions, and test coverage.

### Scope Included in This Branch
1. Admin can ban and unban users from the Users tab.
2. Ban reason and ban timestamp are persisted.
3. Banned users are blocked at login and on protected routes.
4. Ban side effects are applied for operational consistency:
- If the banned user is a requester, their open requests are cancelled and linked active assignments are cleared.
- If the banned user is a volunteer, their active assignment is cancelled, volunteer availability is forced to unavailable, request status is re-synced, and matching is refreshed.
5. Admin Users UI and E2E flow are extended for moderation actions and status visibility.

### Backend Changes
1. Migration
- Added schema support for ban reason and banned timestamp.
- Normalized users.is_banned defaults and null handling.

2. Admin module
- Added ban and unban endpoints:
  - PATCH /api/admin/users/:userId/ban
  - PATCH /api/admin/users/:userId/unban
- Added request validation for optional reason.
- Extended user list payload with banReason and bannedAt.
- Added moderation side-effect orchestration during ban.

3. Auth module
- Login rejects banned users with USER_BANNED (403).
- Protected routes now check current DB user state and reject banned users (403).

4. Availability and request lifecycle integration
- Added volunteer-ban cleanup flow that cancels active assignment, marks volunteer unavailable, syncs request status, and reruns matching.
- Reused existing cancellation and matching flows for deterministic behavior.

### Web Changes
1. Admin Users API client updated with:
- banReason and bannedAt fields.
- ban and unban request functions.

2. Admin Users UI updated with:
- Ban and Unban actions.
- Ban confirmation modal with optional reason input.
- New table columns for Ban Reason and Actions.
- Status and action feedback handling.

3. E2E admin users spec expanded for:
- Ban and unban flow.
- Access blocking while banned.
- Access restoration after unban.

### Testing
1. Updated and added unit tests:
- Admin service moderation orchestration.
- Auth service and controller USER_BANNED handling.
- Related route test mocks updated for auth middleware changes.

2. Updated and added integration tests:
- Admin users ban and unban API behavior.
- Requester-ban and volunteer-ban side effects.
- Auth integration for banned login and banned protected-route behavior.

### Validation Run
1. Backend unit tests passed.
2. Targeted backend integration suites for admin users and auth passed.
3. Web TypeScript check passed.

### Notes
1. This PR is additive and keeps unban behavior simple:
- Unban clears ban fields.
- Volunteer remains unavailable until the user explicitly sets availability again.

Closes #397


kckagancan (changes requested):
Thanks for the follow-up commit. I still see two issues that should be addressed before approval:

1. Admin self-ban/admin-target ban is still not guarded. The backend accepts any `:userId`, and the UI still shows Ban for non-banned rows without checking whether the target is the current admin or another admin. This could allow an admin to ban themselves or lock out other admins. At minimum, self-ban should be blocked at the backend and covered with a test. Ideally, admin accounts should not be bannable unless there is a clear role/permission rule for that.

2. `optionalAuth` still trusts the JWT payload without checking the current DB user state. `requireAuth` now correctly rejects banned users, but optional-auth endpoints can still treat a banned/deleted token as authenticated/admin based on stale JWT claims. This matters for public token-aware endpoints where admin behavior changes the response.

Non-blocking but worth noting: ban side effects are not executed in a single transaction, so a failure during request/assignment cleanup could leave the system partially updated after the user row is already banned.

delibas-rojhat (changes requested):
I found one blocker and one additional behavioral risk:

GET /api/auth/verify-email still issues a fresh access token for a banned user. In backend/src/modules/auth/service.js, verifyUserEmail() calls markEmailVerified() and immediately signs/returns an access token without checking is_banned. That means a banned but not-yet-verified account can still obtain a valid token through the verification flow, which breaks the intended auth enforcement model. Even if protected routes later reject the token with 403, the system is still granting a new authenticated session to a banned account. I think this flow should also return USER_BANNED instead of issuing a token.

The ban side-effect cleanup is not transactional. In backend/src/modules/admin/service.js, the code bans the user first, then cancels requester requests / volunteer assignments, and if a later cleanup step fails it only calls unbanUserById(). That rollback does not restore already-cancelled requests or assignments, so a partial failure can leave the system in a mixed state: the admin sees a failure, the user is unbanned again, but request/assignment side effects may already have been applied. This is a real consistency risk.

Minor UX note, not a blocker:

The ban modal closes before the API request completes, so if the request fails the entered reason is lost. It would be nicer to keep the modal open until the mutation succeeds.

kckagancan (approved):

PR #400, opened 2026-05-01
fix(android): guard location sharing without saved coordinates

### LINKED ISSUE #362: [Bug] Android allows location sharing without saved coordinates
## Summary
  Android can enable location sharing even when there are no saved/shared
  coordinates.

  ## Impact
  The web Privacy & Security page blocks enabling “Share Current Location”
  unless location sharing was already initialized through the profile location
  flow. Android does not have the same guard, so users may enable
  `locationSharingEnabled=true` while the backend has no current coordinates or
  may keep stale coordinates.

  ## Affected files
  - `android/app/src/main/java/com/neph/features/privacysecurity/presentation/
  PrivacySecurityScreen.kt`
  - `android/app/src/main/java/com/neph/features/profile/data/
  ProfileRepository.kt`
  - `web/src/components/feature/settings/PrivacySecurityView.tsx` for reference
  behavior

  ## Suggested fix
  Mirror the web guard on Android:
  - If the user tries to enable location sharing without saved coordinates, show
  an error.
  - Direct the user to update/save current location from the Profile flow first.
  - Alternatively, require a fresh location permission + location capture before
  enabling sharing.

  ## Acceptance criteria
  - Android cannot enable location sharing without valid saved/current
  coordinates.
  - The UX explains how to enable location sharing safely.
  - Web and Android behavior are consistent.
  - Add/update tests for the guard.

erinc00:
This PR fixes Android issue #362 where users could enable location sharing even when there were no valid saved/current coordinates.  
The fix adds a guard to keep Android behavior consistent with web safety expectations.

### Problem
On Android, `Share Current Location` could be turned on in profile flows without a valid coordinate state, causing `locationSharingEnabled=true` with missing or stale coordinates.

### What Changed
1. Repository-level safety guard
- Added a first-time enable guard in profile sync flow.
- If sharing is being enabled from disabled state and there is neither:
  - previously saved shared coordinates, nor
  - freshly captured device coordinates,  
  sync is blocked with a dedicated error.

2. User-facing guidance in Android profile flows
- `EditProfileScreen` now shows a clear error message when the guard blocks unsafe enable.
- `CompleteProfileScreen` now shows the same guidance for consistency.

3. Privacy & Security UX clarification
- Updated Android Privacy & Security screen copy to explain the safe flow:
  - go to Profile,
  - grant permission,
  - capture/save a valid current location first.

4. Tests
- Added/updated unit tests in `ProfileRepositoryLocationPayloadTest` to cover the new guard decision logic:
  - blocks first-time enable without saved/fresh coordinates,
  - allows first-time enable with fresh coordinates,
  - allows first-time enable with previously saved coordinates,
  - does not block when sharing is already enabled.

### Files Changed
- ProfileRepository.kt
- EditProfileScreen.kt
- CompleteProfileScreen.kt
- PrivacySecurityScreen.kt
- ProfileRepositoryLocationPayloadTest.kt

### Acceptance Criteria Mapping
1. Android cannot enable location sharing without valid saved/current coordinates.
- Enforced by repository-level guard.

2. UX explains how to enable sharing safely.
- Added explicit guidance/error messaging in profile/privacy screens.

3. Web and Android behavior are consistent.
- Android now mirrors web’s first-time safety guard intent.

4. Add/update tests for the guard.
- Guard logic covered by new unit tests.

### Validation Notes
Local full Android test execution is environment-blocked by missing `google-services.json` in this workspace setup, but changed files pass editor diagnostics and guard logic is covered by added unit tests.

Closes #362

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -28,6 +28,7 @@ import com.neph.features.profile.data.DeviceLocationProvider
 import com.neph.features.profile.data.LocationData
 import com.neph.features.profile.data.LocationTreeRepository
 import com.neph.features.profile.data.ProfileRepository
+import com.neph.features.profile.data.ProfileRepository.LocationSharingInitializationRequiredException
 import com.neph.features.profile.data.bloodTypeOptions
 import com.neph.features.profile.data.calculateAgeFromDateOfBirth
 import com.neph.features.profile.data.combinePhoneNumber
@@ -250,6 +251,8 @@ fun CompleteProfileScreen(
                 Toast.makeText(context, completionMessage, Toast.LENGTH_LONG).show()
 
                 onComplete()
+            } catch (guardError: LocationSharingInitializationRequiredException) {
+                error = guardError.message ?: "To enable Share Current Location, save a valid current location first."
             } catch (cancellationException: CancellationException) {
                 throw cancellationException
             } catch (errorResponse: ApiException) {
--- a/android/app/src/main/java/com/neph/features/privacysecurity/presentation/PrivacySecurityScreen.kt
+++ b/android/app/src/main/java/com/neph/features/privacysecurity/presentation/PrivacySecurityScreen.kt
@@ -31,11 +31,17 @@ fun PrivacySecurityScreen(
             SectionCard {
                 SectionHeader(
                     title = "Privacy & Security",
-                    subtitle = "Temporary placeholder screen."
+                    subtitle = "Configure privacy safely from your profile flow."
                 )
 
                 Text(
-                    text = "Privacy and security controls will be added in a later step.",
+                    text = "To enable Share Current Location, go to Profile, grant location permission, and save a valid current location first.",
+                    style = MaterialTheme.typography.bodyMedium,
+                    color = MaterialTheme.colorScheme.onSurfaceVariant
+                )
+
+                Text(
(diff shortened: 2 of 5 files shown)


delibas-rojhat (commented):
Thanks for the fix here. I think this PR is moving in the right direction, but I still see one behavior gap that is worth addressing before merge.

1. **The new Android guard still relies on local cached coordinates, not guaranteed valid shared coordinates**

   The new check only looks at whether the previous local profile snapshot has `sharedLatitude/sharedLongitude` or whether a fresh `currentDeviceLocation` exists during the save attempt.

   The problem is that this does not fully guarantee the backend actually has a valid current shared location. If the device cache still contains old coordinates, the guard can pass even when the backend no longer has valid shared coordinates. In that case, Android may still allow `locationSharingEnabled=true`, while the request does not send fresh coordinates in the same save flow.

   So the PR improves the situation, but I do not think it completely closes the issue described in the ticket yet. The remaining edge case is:

   - local cache has coordinates
   - backend does not have valid current shared coordinates
   - Android allows enabling sharing
   - no fresh coordinates are sent in that save attempt

   That means the system can still end up with location sharing enabled without a guaranteed valid current location on the backend.

   I think the safer fix would be to require one of these when enabling sharing:
   - a fresh location capture in the same flow, or
   - a backend-confirmed saved/shared coordinate state, rather than only the local cached profile state

Because of that, I would be careful about treating this as fully aligned with the issue’s acceptance criteria yet, especially the “Web and Android behavior are consistent” and “Android cannot enable location sharing without valid saved/current coordinates” parts.

kckagancan (approved):

PR #401, opened 2026-05-01
Align profile name/DOB flow across web-android-backend

### LINKED ISSUE #287: [FEATURE][Core Flow] Revise auth, verification, and profile flows
**Description**
Authentication, verification, and profile-related flows should be reviewed end-to-end and revised where necessary. The goal is to improve correctness, consistency, validation, edge-case handling, and alignment between web, mobile, and backend behavior.

**Tasks**
- Review login flow
- Review signup flow
- Review verification flow
- Review guest vs authenticated behavior where applicable
- Review profile retrieval/update flow
- Fix validation inconsistencies
- Fix UX issues and flow edge cases
- Align backend, web, and mobile expectations/responses
- Re-test full auth/profile lifecycle

**Acceptance Criteria**
- Login, signup, and verification flows behave consistently
- Profile flow is stable and coherent across platforms
- Validation and error handling are improved
- Backend/web/mobile behavior is aligned

**Dependencies**
- Cleanup and existing bug fixes recommended first

erinc00:
This PR aligns the profile data model and UX across backend, web, and Android by moving to first name + last name + date of birth as the primary flow, while preserving backward compatibility where needed.

## What Changed
1. Backend profile schema and mapping were extended for date of birth.
2. Web profile and complete-profile flows were updated from full name/age to first name, last name, and date of birth.
3. Android profile model, repository mapping, and profile screens were updated to the same first name/last name/date of birth structure.
4. Profession options were aligned by removing Volunteer from selectable options.
5. Profile/location UX was refined:
   - Profile avatar initials now prioritize first name + last name.
   - Profile medical upload UI was removed while keeping chronic diseases/allergies text fields.
   - Use Current Location button alignment was adjusted in the location picker.
6. Regression-safe backend test updates were added for date of birth behavior and mapping.

## Why
The previous state had cross-platform inconsistencies in profile data handling, causing UX mismatch and backend/profile read issues when schema and client expectations diverged. This PR standardizes the flow and reduces those integration risks.

## Validation
1. Focused backend unit tests for profile service/validator pass.
2. Migration for date_of_birth was applied to align DB schema with backend queries.
3. Web and Android changes were implemented with minimal scope to avoid unrelated behavior changes.

## Notes
1. The approach keeps compatibility fallbacks where possible to avoid breaking existing users/data.
2. Out-of-scope auth/moderation hardening items were not included in this PR.

Closes #287

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -29,9 +29,12 @@ import com.neph.features.profile.data.LocationData
 import com.neph.features.profile.data.LocationTreeRepository
 import com.neph.features.profile.data.ProfileRepository
 import com.neph.features.profile.data.bloodTypeOptions
+import com.neph.features.profile.data.calculateAgeFromDateOfBirth
 import com.neph.features.profile.data.combinePhoneNumber
+import com.neph.features.profile.data.composeFullName
 import com.neph.features.profile.data.expertiseOptionsFor
 import com.neph.features.profile.data.locationData
+import com.neph.features.profile.data.normalizeDateOfBirth
 import com.neph.features.profile.data.normalizePhoneParts
 import com.neph.features.profile.data.parseListField
 import com.neph.features.profile.data.professionOptionsFor
@@ -53,6 +56,9 @@ import com.neph.ui.map.SharedCoordinatesMapAction
 import com.neph.ui.theme.LocalNephSpacing
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.launch
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
 
 @Composable
 fun CompleteProfileScreen(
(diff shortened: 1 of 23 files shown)


delibas-rojhat (changes requested):
Thanks for the update here. I reviewed the PR and I think the main goal makes sense, but I still see two issues that should be addressed before merge:

1. **Existing `Volunteer` professions can be lost on the web profile flow**

   This PR is framed as restricting expertise options to `First Aid`, but it also changes how the existing `Volunteer` profession is handled on web. Right now, if a user already has `Volunteer` stored as their profession, the profile mapping converts that into an empty profession value, and the profession dropdown no longer includes `Volunteer` as a valid option.

   The problem is that this turns an existing stored value into what looks like “no profession selected”, and then a normal profile save can send `null` back to the backend. In practice, that means a user can lose their profession simply by editing some unrelated field and pressing save.

   That feels like unintended data loss, and it also goes beyond the stated scope of limiting expertise options.

2. **Android does not handle existing `Volunteer` professions consistently**

   A similar issue exists on Android. `Volunteer` was removed from the profession options there as well, but the edit flow still binds the dropdown to the stored profession value.

   So if a user already has `Volunteer` in their profile, the screen can end up in an inconsistent state where the current stored value is no longer actually present in the dropdown options. Even if this does not immediately crash anything, it creates a confusing UX and makes it easier to accidentally overwrite valid existing data.

   I think the safer approach would be:
   - restrict *new* selections if that is the intended product decision
   - but still preserve and render already-stored legacy values correctly until they are explicitly migrated or replaced

erinc00:
> Thanks for the update here. I reviewed the PR and I think the main goal makes sense, but I still see two issues that should be addressed before merge:
> 
> 1. **Existing `Volunteer` professions can be lost on the web profile flow**
>    This PR is framed as restricting expertise options to `First Aid`, but it also changes how the existing `Volunteer` profession is handled on web. Right now, if a user already has `Volunteer` stored as their profession, the profile mapping converts that into an empty profession value, and the profession dropdown no longer includes `Volunteer` as a valid option.
>    The problem is that this turns an existing stored value into what looks like “no profession selected”, and then a normal profile save can send `null` back to the backend. In practice, that means a user can lose their profession simply by editing some unrelated field and pressing save.
>    That feels like unintended data loss, and it also goes beyond the stated scope of limiting expertise options.
> 2. **Android does not handle existing `Volunteer` professions consistently**
>    A similar issue exists on Android. `Volunteer` was removed from the profession options there as well, but the edit flow still binds the dropdown to the stored profession value.
>    So if a user already has `Volunteer` in their profile, the screen can end up in an inconsistent state where the current stored value is no longer actually present in the dropdown options. Even if this does not immediately crash anything, it creates a confusing UX and makes it easier to accidentally overwrite valid existing data.
>    I think the safer approach would be:
>    
>    * restrict _new_ selections if that is the intended product decision
>    * but still preserve and render already-stored legacy values correctly until they are explicitly migrated or replaced

Thanks, this concern is valid in a migration scenario. For this release we are launching with a clean database and no legacy profile migration, so existing Volunteer records are not expected. We intentionally restrict selectable options for new data only.  Given these release assumptions, I believe this specific review concern is not applicable to this PR.

kckagancan (changes requested):
Thanks for the work here. I still see a few issues before approval:

1. Web Profile edit flow does not appear to let an existing user update first name / last name / phone anymore. Complete Profile handles first/last name, but `ProfileView` save flow seems to patch physical/health/location/privacy/profession/expertise only, not `/profiles/me`. Since this PR is meant to align the web profile flow around first name + last name + DOB, existing profile editing should also support those fields or this should be explicitly out of scope.

2. Backend `dateOfBirth` validation uses `Date.parse` and normalization, which is not strict enough for a profile data contract. Android uses strict `yyyy-MM-dd` parsing, but backend may normalize invalid calendar dates instead of rejecting them. For cross-platform consistency, backend should use strict `YYYY-MM-DD` validation and reject invalid dates like non-existent calendar days.

3. The existing `Volunteer` profession concern depends on release assumptions. If the database is guaranteed clean, the current approach is acceptable. If any existing profile data can contain `Volunteer`, this can still lead to data loss or inconsistent dropdown state.

Non-blocking: backend still accepts arbitrary expertise areas while web/android only expose `First Aid`. That may be intentional for backward compatibility, but the API/client contract should be clear.

kckagancan (approved):

PR #418, opened 2026-05-03
fix: add in-app OSM map picker for location selection for mobile

### LINKED ISSUE #417: [FIX] In-app Map Picker
## Description

Mobile currently supports location-related flows, but selected locations cannot be chosen directly from an in-app map.

The existing shared map integration only opens the selected address or coordinates in an external map app. This is useful for viewing, but it does not let the user pick a point on the map and apply that selection back to the form.

This creates a UX gap in mobile location flows, especially in Profile / Complete Profile / Request Help screens where users expect to choose their location visually.

## Current Behavior

- Users can select country / city / district / neighborhood manually through dropdowns.
- Users can open selected locations in an external map app.
- Users can share current GPS location when permission is granted.
- Guest Request Help can use current location autofill.
- However, users cannot pick a location from a map and apply it to the form.

## Expected Behavior

Users should be able to open an in-app map picker, select a location point, and apply it back to the related mobile form.

The selected map location should update location fields as much as possible using reverse geocoding and should not break the existing manual dropdown flow.

## Scope

Android/mobile only.

Affected flows:

- Complete Profile location section
- Edit Profile location section
- Request Help location section

Out of scope:

- Web map picker
- Backend schema changes unless absolutely required
- Replacing manual dropdowns entirely
- Redesigning Gathering Areas
- Changing Share Current Location privacy behavior
- Changing help request matching logic

## Tasks

- Add a reusable Android in-app map picker flow/component.
- Allow users to open the picker from mobile location sections.
- Let users select a point on the map.
- Use the existing reverse geocoding flow where possible to resolve:
  - country
  - city
  - district
  - neighborhood
  - extra address / address description
- Apply resolved location fields back to the form.
- Keep manual country / city / district / neighborhood dropdowns available as fallback.
- Preserve existing current-location permission behavior.
- Preserve existing external “Open in Map” behavior if still useful.
- Reuse shared map/location code instead of duplicating logic across screens.
- Add focused tests for map-picker result mapping and form state updates where practical.

## Acceptance Criteria

- Complete Profile allows selecting a location from an in-app map and applies the result to the location fields.
- Edit Profile allows selecting a location from an in-app map and applies the result to the location fields.
- Request Help allows selecting a location from an in-app map and applies the result to the request location fields.
- Manual dropdown-based location selection still works after the change.
- Current-location sharing and guest current-location autofill still work after the change.
- If reverse geocoding fails or returns partial data, the user can still manually complete the fields.
- No unrelated backend, web, matching, or UI redesign changes are introduced.
- Android build/tests pass.

## Notes

This issue fixes the gap where mobile has location support but does not yet support actual map-based location picking. Existing external map opening should not be confused with in-app location selection.

kckagancan:
## Summary

This PR adds a reusable Android in-app map picker for mobile location selection.

Previously, mobile location flows supported dropdown-based location selection, current-location sharing, guest current-location autofill, and external “Open in Map” actions, but users could not select a point directly inside the app and apply it back to the form.

## What changed

- Added a reusable in-app OpenStreetMap/Leaflet-based map picker dialog.
- Added tap-to-pin location selection with confirm/cancel behavior.
- Added reverse-geocode mapping from selected coordinates into existing location fields where possible:
  - country
  - city
  - district
  - neighborhood
  - extra address / address description
- Added `Select Location on Map` action to:
  - Complete Profile
  - Edit Profile
  - Request Help
- Kept the existing manual dropdown location flow available after map selection.
- Added focused mapping tests for map picker result handling.

## Important behavior notes

- This PR does **not** use Google Maps SDK.
- This PR does **not** add Google Maps API keys.
- This PR does **not** add Play Services Maps for the picker.
- The picker is in-app and OpenStreetMap-based.
- Existing external `Open in Map` actions are preserved.
- Existing current-location sharing behavior is preserved.
- Existing guest Request Help current-location autofill behavior is preserved.
- If reverse geocoding returns partial data, only safe/resolved fields are applied and the user can still edit location manually.

## Scope notes

- Android/mobile only.
- No web changes.
- No backend schema changes.
- No help-request matching changes.
- No gathering-areas changes.
- No broad UI redesign.

## Validation

- `./gradlew :app:testE2eUnitTest`
- `./gradlew :app:assembleE2eAndroidTest`

## Limitations / follow-up

- The map picker uses network-loaded OpenStreetMap tiles through WebView, so map tile rendering requires network access.
- Selected coordinates are currently used for reverse geocoding and form-field population. Precise coordinate persistence depends on the existing backend contract for each flow.

Closes #417

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -42,8 +42,10 @@ import com.neph.features.profile.data.professionOptionsFor
 import com.neph.features.profile.data.sanitizeDecimalInput
 import com.neph.features.profile.data.splitFullName
 import com.neph.features.profile.data.toEditableString
+import com.neph.features.requesthelp.data.RequestHelpRepository
 import com.neph.features.profile.presentation.components.GenderSelector
 import com.neph.features.profile.presentation.components.LocationSelector
+import com.neph.ui.components.buttons.SecondaryButton
 import com.neph.ui.components.display.HelperText
 import com.neph.ui.components.display.SaveActionBar
 import com.neph.ui.components.inputs.AppDropdown
@@ -52,8 +54,12 @@ import com.neph.ui.components.inputs.AppTextField
 import com.neph.ui.components.selection.AppCheckbox
 import com.neph.ui.components.selection.AppToggleSwitch
 import com.neph.ui.layout.AuthScaffold
+import com.neph.ui.map.MapPickerDialog
+import com.neph.ui.map.MapPickerSelection
 import com.neph.ui.map.LocationSelectionMapAction
 import com.neph.ui.map.SharedCoordinatesMapAction
+import com.neph.ui.map.formatMapCoordinate
+import com.neph.ui.map.resolveMapPickerLocationUpdate
 import com.neph.ui.theme.LocalNephSpacing
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.launch
@@ -105,6 +111,9 @@ fun CompleteProfileScreen(
     var error by rememberSaveable { mutableStateOf("") }
     var info by rememberSaveable { mutableStateOf("") }
     var mapActionInfo by rememberSaveable { mutableStateOf("") }
+    var mapPickerOpen by rememberSaveable { mutableStateOf(false) }
+    var mapPickerLoading by rememberSaveable { mutableStateOf(false) }
+    var mapPickerSelection by remember { mutableStateOf<MapPickerSelection?>(null) }
     var availableLocationData by remember { mutableStateOf<LocationData>(locationData) }
     var locationLoading by remember { mutableStateOf(true) }
     var locationInfo by rememberSaveable { mutableStateOf("") }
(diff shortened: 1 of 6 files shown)


mehmetcangurbuz08 (changes requested):
Great work overall, especially on making the map picker reusable across 3 flows and applying reverse-geocode mapping in a controlled way. Keeping the manual dropdown flow available after map selection is also a strong UX decision.

That said, I recommend Request Changes before merge for these points:

Security: WebView + JS bridge + remote scripts
addJavascriptInterface is used while Leaflet/CSS are loaded from CDN (unpkg), which increases risk.
Please move Leaflet assets to local bundled files (or enforce a very strict loading policy).
Refs: MapPickerDialog.kt (line 159), MapPickerDialog.kt (line 211), MapPickerDialog.kt (line 212)

Missing WebView navigation restrictions
WebViewClient() is left as default; navigation should be restricted with an allowlist / shouldOverrideUrlLoading so only expected content can load.
Ref: MapPickerDialog.kt (line 158)

Encoding issue in tests
Test data includes mojibake like "BeÅŸiktaÅŸ"; this should be corrected to proper UTF-8 ("Beşiktaş") to keep test validity against real data.
Refs: MapPickerMappingTest.kt (line 21), MapPickerMappingTest.kt (line 105)

mehmetcangurbuz08 (approved):
Looks good. Merging.

PR #422, opened 2026-05-03
fix: shown priority and urgency info to user

### LINKED ISSUE #409: [Bug] Priority & urgency info of the help request shown to the user requesting help
It does not make sense to show the priority & urgency info of the help request to the very same user that requested help.

kckagancan:
## Summary

This PR hides the backend-computed urgency and priority values from the requester-facing **My Help Requests** screen on Android.

## Reason

The urgency and priority levels are calculated by the backend as operational/triage values. Since these values are meant to support internal response prioritization, assignment, volunteer, or admin workflows, the user who created the help request does not need to see them on their own request details screen.

The related requirement was a bit ambiguous because it mentions showing help request details including urgency level. This seems to have been interpreted as requester-facing visibility, but in this case the displayed urgency/priority values are backend-computed operational fields rather than user-entered request details.

## Changes

- Removed `Urgency: ...` from the requester-facing help request card.
- Removed `Priority: ...` from the requester-facing help request card.
- Kept the underlying data/model mapping unchanged so these values can still be used in backend, volunteer, admin, or assignment-related flows.

## Scope

- Android only.
- UI-only change.
- No backend, database, API, or sync behavior changes.

## Related Issue

Closes #409

### DIFF
--- a/android/app/src/main/java/com/neph/features/myhelprequests/presentation/MyHelpRequestsScreen.kt
+++ b/android/app/src/main/java/com/neph/features/myhelprequests/presentation/MyHelpRequestsScreen.kt
@@ -141,16 +141,16 @@ fun MyHelpRequestsScreen(
                         }
                     }
 
-                        item {
-                            SectionHeader(
-                                title = "Current Request",
-                                subtitle = if (isAuthenticated) {
-                                    "Your latest active help request is shown first."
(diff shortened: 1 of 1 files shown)


erinc00 (approved):
Looks good now, merging.

PR #423, opened 2026-05-03
chore(web): remove switch account from top navbar menu

erinc00:
Removed the Switch Account option from the menu since it duplicated the logout flow.


kckagancan (approved):
Looks fine. Merging.

Issue #383, opened 2026-04-30
[Feature] Add demo-ready seed migrations with meaningful NEPH data


AlperKartkaya:
## Summary

  Add migration/seed support for meaningful demo data so the deployed app is not empty
  during presentations, evaluation, or customer demos.

  The goal is to populate the system with realistic Neighborhood Emergency Preparedness
  Hub data: users, profiles, helpers, help requests, admin-visible emergency activity,
  notifications, and announcements.

  ## Problem

  Currently, after deployment or migration, the app may have little or no useful data.
  This makes demo presentations weaker because many pages appear empty or do not clearly
  show the value of the system.

  We need meaningful, realistic seed data that demonstrates the core purpose of the app
  without requiring manual database edits.

  ## Scope

  Create an idempotent demo-data migration or seed script that can populate the app with
  realistic sample data.

  Suggested data to include:

  - Demo admin account
  - Demo resident/requester accounts
  - Demo helper/volunteer accounts
  - Completed user profiles with realistic Istanbul locations
  - Helper availability and expertise areas
  - Active help requests with different urgency levels
  - Resolved or historical help requests
  - Emergency announcements
  - In-app notifications
  - Admin dashboard data that makes overview/history/insights pages useful
  - Optional fallback/demo gathering areas or shelter-style locations if supported by
  the schema

  ## Example Demo Data

  Use realistic but clearly non-real emergency-preparedness content, such as:

  - Locations:
    - Istanbul / Beşiktaş / Levazım
    - Istanbul / Kadıköy / Moda
    - Istanbul / Şişli / Mecidiyeköy
  - Help request categories:
    - Medical assistance
    - Water/food need
    - Mobility support
    - Search and rescue support
  - Risk flags:
    - Injury
    - Elderly person affected
    - Mobility limitation
    - Urgent medication need
  - Announcements:
    - “Temporary gathering area opened in Beşiktaş”
    - “Water distribution available near Kadıköy”
    - “Avoid damaged road near Şişli”

  All demo content should be clearly marked as demo/test data where appropriate.

  ## Implementation Requirements

  - [ ] Add a repeatable migration or seed script for demo data.
  - [ ] Script must be idempotent: running it multiple times must not create duplicate
  records.
  - [ ] Script must not destroy or overwrite real user data.
  - [ ] Demo records should use stable identifiers, emails, or tags so they can be
  updated safely.
  - [ ] Demo data should be realistic and useful for showing the product’s purpose.
[...]

PR #424, opened 2026-05-03
feat: add idempotent demo data seed script

gulcetahtasiz:
## Summary
- Added an idempotent demo data seed script for the NEPH backend
- Seed includes demo users, profiles, volunteers, help requests, assignments, announcements, and notifications
- Added README instructions for running the seed locally or on EC2 with Docker Compose

## Notes
- Demo data is guarded by `ENABLE_DEMO_SEED=true`
- Running the script multiple times skips existing demo records and does not delete real data

## Test
- Ran `node --check scripts/seed-demo-data.js`

## Demo accounts:

| Role | Email | Password |
| --- | --- | --- |
| Admin | admin_demo@neph.test | DemoPass123! |
| Requester | requester_ayse@neph.test | DemoPass123! |
| Requester | requester_mert@neph.test | DemoPass123! |
| Volunteer | volunteer_elif@neph.test | DemoPass123! |
| Volunteer | volunteer_can@neph.test | DemoPass123! |

Closes [#383](https://github.com/bounswe/bounswe2026group6/issues/383)

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,29 @@ Notes:
 - the backend waits for PostgreSQL to become healthy before starting
 - the web container is configured to talk to the backend in the local Docker setup
 
+### Demo data seed
+
+After migrations have run, you can populate the app with clearly marked demo data for presentations and evaluation. The seed is idempotent, so running it again skips existing demo records instead of creating duplicates.
+
+For the Docker Compose setup, including EC2 deployments that run this repository with Compose:
+
+```bash
+docker compose exec -e ENABLE_DEMO_SEED=true backend node scripts/seed-demo-data.js
+```
+
+For a backend process running directly on the host:
+
+```bash
+cd backend
+ENABLE_DEMO_SEED=true node scripts/seed-demo-data.js
+```
+
+Demo login password for `*@neph.test` users:
+
+```text
+DemoPass123!
+```
+
 ### Optional non-Docker development
 
 If you want to run modules directly on your machine instead of Docker Compose, use the module-specific setup inside each folder:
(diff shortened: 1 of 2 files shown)


kckagancan (changes requested):
Thanks @gulcetahtasiz , this is a solid demo seed foundation and the scope mostly matches #383.

Before approval, I think two things should be tightened:

1. The seed should populate minimal physical_info, health_info, and privacy_settings rows for demo users. The issue asks for completed demo profiles and preparedness information, but currently profiles mostly get identity + location, while health/physical/privacy sections can remain empty.

2. The idempotency logic should handle pre-existing demo emails/users/profiles more safely. Right now, if a demo email already exists with a different user_id, the user insert is skipped but child records still reference the hard-coded demo_user_* id, which can fail with FK errors. Similarly, if a profile exists with a different profile_id, later location/expertise inserts still use the deterministic demo_profile_* id. Please either fetch/reuse existing IDs or fail early with a clear message.

Also, please confirm the script was run against a migrated local Docker DB twice and that the second run only skipped existing demo records. `node --check` verifies syntax only, not SQL/schema/idempotency behavior.

gulcetahtasiz:
Thanks for the review. I addressed both points:

- Added minimal `physical_info`, `health_info`, and `privacy_settings` rows for all demo users.
- Updated the idempotency logic to fetch and reuse existing `user_id`, `profile_id`, `volunteer_id`, and `admin_id` values instead of assuming hard-coded IDs.
- Added clear conflict errors when deterministic demo IDs are already used by unrelated records.
- Verified against a migrated local Docker DB by running the seed twice. The second run produced `total: inserted 0, skipped 48`.

Could you re-check? @kckagancan

kckagancan (approved):
Thanks for the fixes. Looks good now. Merging

Issue #429, opened 2026-05-04
[Bug] Fix Profile Residential Location Selection and Map UX


kckagancan:
## Summary

Android profile location editing currently mixes residential/profile location, current/shared GPS coordinates, map selection, and location-sharing behavior in the same UI. Manual location selection may also fail because dependent dropdown resets are triggered in both the parent screen and the `LocationSelector` component.

This issue fixes the profile location UI and makes it consistent with the new residential vs operational location model.

## Problem

Profile location should represent the user's home/neighborhood/community location.

However, the current Edit Profile / Complete Profile screens include:

- Country/city/district/neighborhood dropdowns
- `Select Location on Map`
- selected coordinate helper text
- external map action for selected residential fields
- external map action for shared/current coordinates
- `Share Current Location` toggle
- current GPS capture during profile save when sharing is enabled

This creates confusing behavior:

- The user may think map-selected coordinates are saved as their home location.
- The app may instead save current GPS as shared coordinates.
- Current GPS may be confused with residential/profile location.
- Manual dropdown selection may not work reliably.

## Root Cause

`LocationSelector` calls dependent reset callbacks internally, while parent screens also reset dependent fields.

Example behavior:

- Child calls `onCountryChange(it)`
- Then child also calls `onCityChange("")`
- Parent `onCountryChange` already resets city/district/neighborhood
- Parent `onCityChange("")` may run with stale parent state and overwrite the selected country

This can cause selected values to reset unexpectedly.

### Fix location dropdown state updates

- Ensure country/city/district/neighborhood selection works reliably.
- Keep dependent reset logic in only one place.
- Preferred approach:
  - `LocationSelector` should only emit the selected field change.
  - Parent screen should handle dependent resets.
- Apply the fix wherever `LocationSelector` is used:
  - Complete Profile
  - Edit Profile
  - Request Help, if affected

### Clarify residential/profile location UI

- Rename the profile section from generic `Location` to something like:
  - `Home / Neighborhood Location`
  - or `Residential Location`
- Add helper copy:
  - "This is your home/neighborhood location. It is used for community and nearby-neighbor features. It is not automatically changed by GPS."

### Separate current/shared location from profile location

- Remove `Share Current Location` from the residential profile location card, or move it into a clearly separate section.
- Profile save must not automatically capture current GPS.
- Profile save must only save manually selected residential fields.

### Fix map picker semantics

- `Select Location on Map` should mean selecting the residential/home/neighborhood location.
[...]

PR #442, opened 2026-05-04
fix: Android profile residential location handling

gulcetahtasiz:
## Summary
- Makes `LocationSelector` stateless so dependent dropdown resets are handled only by parent screens.
- Separates residential profile location from current/shared GPS location in Complete Profile and Edit Profile.
- Updates profile location UI copy to clarify home/neighborhood semantics.
- Prevents profile save from implicitly capturing current GPS.
- Updates map picker so opening the map only centers it; users must place and confirm a pin before fields update.

## Verification
- `./gradlew :app:compileDebugKotlin -x processDebugNavigationResources`
- `./gradlew :app:assembleDebug`

Closes [#429](https://github.com/bounswe/bounswe2026group6/issues/429)

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -1,8 +1,6 @@
 package com.neph.features.auth.presentation
 
 import android.widget.Toast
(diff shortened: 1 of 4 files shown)


kckagancan (approved):
Thanks for the bug fix. Looks good now. Merging.

PR #443, opened 2026-05-04
fix: contextual location permission ux

### LINKED ISSUE #432: [Feature][Android] Implement Contextual Foreground Location Permission UX
## Summary

Implement a consistent Android foreground location permission flow.

NEPH should **not** ask for location permission repeatedly. If permission is already granted, the app should use current location directly. If permission is missing, it should request permission only after the user starts a location-dependent action.

No background / always location permission should be requested in MVP.

## Core Rule

```text
if foreground location permission is already granted:
    use current location directly
else:
    request permission only after a user-triggered location action
```

Once the user grants "Allow while using the app", NEPH can reuse that permission in later foreground flows unless Android/user settings revoke it or the user selected a temporary permission mode.

## Location Permission Can Be Requested Only For

- **Request Help / Emergency Mode:** when the user taps `Use Current Location`.
- **Available to Help ON:** when the user wants to become assignable by nearby matching.
- **Safety Check-in:** only if the user chooses to share location.
- **Profile map picker:** only if the user taps `Center on my location`.
- **Home optional card:** only if the user taps an `Enable Location` card.

Do not request permission automatically on app launch, app resume, or normal profile editing.

### App launch / resume

- If permission is already granted, the app may refresh operational location silently.
- If permission is missing, do nothing.
- Never show the system permission dialog on launch/resume.
- Never update profile/residential location from this flow.

### Home optional card

- Show only when foreground location permission is missing.
- Tapping it requests permission.
- Granting permission must not update profile, create a request, enable availability, or trigger assignment.

### Request Help

- If permission is granted, use current location directly.
- If permission is missing, ask when the user taps `Use Current Location`.
- If denied, allow manual emergency location entry.
- Works for both guest and authenticated users.

### Available to Help

- If permission is granted, capture current location directly.
- If permission is missing, ask when user turns availability ON.
- If denied or location is unavailable, do not make the volunteer auto-assignable.
- Turning availability OFF must not require location.

### Safety Check-in

- Marking safe does not require location.
- Ask permission only if the user chooses to share location.
- If denied, mark safe without location.

### Profile map picker

- Normal profile editing must not request location permission.
- The map may have a `Center on my location` action.
- That action checks permission and requests it only if missing.
- Permission grant only centers the map; it must not save or modify profile location.
- Profile save must not capture current GPS by itself.

## Tasks

- Add a shared foreground location permission helper/composable.
- Ensure location-dependent flows check permission before requesting it.
- Add optional Home `Enable Location` card.
- Integrate the helper into:
  - Request Help
  - Available to Help
  - Safety Check-in
  - Profile map picker
- Handle denied/revoked permission with clear fallback messages.
- Do not request `ACCESS_BACKGROUND_LOCATION`.

## Acceptance Criteria

- Permission is not requested repeatedly after it has already been granted.
- Permission is requested only after a user-triggered location action.
- App launch/resume never shows the permission dialog.
- Normal profile editing never shows the permission dialog.
- Request Help supports current location for guest and authenticated users.
- Available to Help ON requires usable current location before becoming auto-assignable.
- Safety Check-in can proceed without location if permission is denied.
- Profile map permission only centers the map and does not update profile location.
- The app does not request background / always location permission.

kckagancan:
## Summary

This PR implements contextual foreground location permission handling on Android.

Location permission is now requested only after the user starts a location-dependent action, instead of being requested automatically during unrelated flows.

## Changes

- Added/reused shared foreground location permission handling for Android flows
- Added an optional Home card to enable foreground location permission
- Updated Request Help current-location flow to request permission only after `Use Current Location`
- Updated Available to Help flow so turning availability ON requires usable current location
- Updated Safety Check-in so location permission is requested only when the user chooses to share location
- Updated profile map behavior so current-location permission only centers the map and does not save profile location
- Added fallback messages for denied or unavailable location states

## Important Behavior

- The app does not request location permission on launch or resume
- Normal profile editing does not request location permission
- The app does not request background / always location permission
- Granting permission does not automatically update profile location, create a request, enable availability, or trigger assignment

## Testing

- Verified relevant Android build/checks
- Verified location-dependent flows where applicable

Closes #432

### DIFF
--- a/android/app/src/main/java/com/neph/MainActivity.kt
+++ b/android/app/src/main/java/com/neph/MainActivity.kt
@@ -10,21 +10,18 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
-import androidx.lifecycle.lifecycleScope
 import androidx.navigation.compose.rememberNavController
 import com.neph.core.NephAppContext
 import com.neph.core.database.NephDatabaseProvider
 import com.neph.core.sync.OfflineSyncScheduler
 import com.neph.features.availability.data.AvailabilityRepository
 import com.neph.features.auth.data.AuthSessionStore
-import com.neph.features.profile.data.AppLaunchLocationUpdater
 import com.neph.features.profile.data.ProfileRepository
 import com.neph.features.notifications.data.PushTokenSync
 import com.neph.features.requesthelp.data.RequestHelpRepository
 import com.neph.navigation.AppNavGraph
 import com.neph.navigation.Routes
 import com.neph.ui.theme.NephTheme
-import kotlinx.coroutines.launch
 
 class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -39,9 +36,6 @@ class MainActivity : ComponentActivity() {
         PushTokenSync.syncCurrentToken()
         OfflineSyncScheduler.schedulePeriodicSync(applicationContext)
         OfflineSyncScheduler.enqueueSync(applicationContext, reason = "app-start")
-        lifecycleScope.launch {
-            AppLaunchLocationUpdater.updateOnAppLaunch(applicationContext)
-        }
         setContent {
             NephApp()
         }
--- a/android/app/src/main/java/com/neph/features/availability/data/AvailabilityRepository.kt
+++ b/android/app/src/main/java/com/neph/features/availability/data/AvailabilityRepository.kt
@@ -12,6 +12,7 @@ import com.neph.core.sync.OfflineSyncScheduler
 import com.neph.core.sync.SyncEntityType
 import com.neph.core.sync.SyncOperationType
 import com.neph.core.sync.SyncStatus
+import com.neph.features.profile.data.CurrentDeviceLocation
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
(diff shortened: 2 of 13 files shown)


gulcetahtasiz (changes requested):
Thanks, this is moving the app in the right direction. The contextual permission model looks good overall, especially removing app-launch location capture and only asking permission after explicit location-dependent actions.

I found two issues before approval:

1. Map picker keeps a stale selected pin after “Center on my location”

In `MapPickerDialog`, the displayed map center can change through `centerLatitude/centerLongitude`, but `selection` is remembered only from `initialLatitude/initialLongitude`.

That means this flow is possible:
- user opens profile map picker
- taps a point, so `selection` is set
- taps “Center on my location”
- map recenters, but `selection` still points to the old pin
- “Use Selected Location” remains enabled
- user can confirm an old/invisible coordinate

Since centering should not select/save anything by itself, could we clear `selection` whenever the center coordinates change, or include `centerLatitude/centerLongitude` in the `remember(...)` key so the user must place a new pin after centering?

2. Availability sync coordinates need backend range validation

The sync endpoint now persists latest availability coordinates into volunteer last-known coordinates, but validation only checks that `latitude` and `longitude` are numbers. Out-of-range finite values can pass validation and then fail the DB coordinate CHECK constraint, returning a 500 instead of a 400.

Could we validate:
- latitude between `-90` and `90`
- longitude between `-180` and `180`
- latitude/longitude must be provided together

After these are fixed, this should be good to approve.

gulcetahtasiz (approved):
Thanks for the fixes. Looks good to me now. Merging

Issue #431, opened 2026-05-04
[Feature][Android] Add Operational Location Repository and App Launch/Resume Refresh


kckagancan:
## Summary

Add Android-side operational location storage and sync. App launch/resume should refresh operational location opportunistically only if foreground location permission already exists.

This replaces the current behavior where app launch updates profile location.

## Problem

Current app launch location update goes through the profile flow and patches `/profiles/me/location`. That is incorrect because profile location should represent residential/home/neighborhood location.

App launch should update operational location only.

### App launch / app resume

On app launch or resume:

```text
if foreground location permission is already granted:
    capture current device location
    store it in local operational location cache
    if user is authenticated:
        sync it to backend operational location endpoint
    if user is guest:
        keep it local only
else:
    do nothing
```

Important:

- Do not request permission automatically on app launch.
- Do not block app launch.
- Do not update profile/residential location.
- Do not create help requests.
- Do not enable volunteer availability.
- Do not show noisy errors for GPS/network failure.

## Local Model

Create local operational location storage containing:

- latitude
- longitude
- accuracyMeters
- source
- capturedAt
- updatedAtEpochMillis
- syncStatus, if authenticated backend sync is pending/failed

This may be implemented through Room or SharedPreferences depending on existing local architecture, but Room is preferred if sync status is needed.

## Sync Rules

- Authenticated user:
  - sync to `/operational-location/me` when online
  - queue or retry if offline
- Guest user:
  - store locally
  - do not call authenticated backend endpoint
  - use local operational location only for guest Request Help flow

## Freshness

Define constants:

```text
OPERATIONAL_LOCATION_REFRESH_THROTTLE = 30 minutes
```

App launch/resume should avoid repeated refresh if the local operational location is already fresh enough.

## Tasks

- Create `OperationalLocationRepository`.
- Create local model/cache for operational location.
- Replace `AppLaunchLocationUpdater` behavior with operational location update.
- Stop calling `ProfileRepository.syncLocationOnLaunch()` from app launch.
- Add app resume refresh if appropriate and not disruptive.
- Sync authenticated operational location to backend endpoint.
- Keep guest operational location local-only.
- Add unit tests for freshness/throttle decisions where possible.

## Acceptance Criteria

- App launch no longer patches `/profiles/me/location`.
- App launch does not request location permission.
- If permission is already granted, app launch can refresh operational location.
- Operational location refresh failure does not block app start.
- Guest users do not call authenticated operational location API.
[...]

PR #445, opened 2026-05-05
feat: add Android operational location refresh

gulcetahtasiz:
## Summary
Adds Android-side operational location storage and silent launch/resume refresh, separating runtime GPS location from residential profile location.

## Changes
- Adds local Room storage for latest operational location.
- Adds `OperationalLocationRepository`.
- Adds `OperationalLocationUpdater`.
- Refreshes operational location on app launch/resume only if foreground location permission already exists.
- Syncs authenticated users to `/operational-location/me`.
- Keeps guest operational location local-only.
- Removes app-launch profile location sync.
- Removes `ProfileRepository.syncLocationOnLaunch()`.

## Behavior
- App launch does not request location permission.
- App launch does not patch `/profiles/me/location`.
- Residential/profile location remains unchanged.
- GPS/network failures are silent and do not block startup.
- Refresh is throttled to 30 minutes.
- Failed authenticated sync is kept as pending for retry on a later launch/resume.

## Verification
- `./gradlew :app:compileDebugKotlin`

## Notes
- This PR does not modify backend APIs.
- Guest Request Help can consume the local operational location cache in a follow-up if needed.

Closes [#431](https://github.com/bounswe/bounswe2026group6/issues/431)


kckagancan (changes requested):
Thanks for the PR. The overall direction is correct and it fixes the most important architectural problem However, there are a few behavioral correctness issues, especially around freshness/throttling.

### 1. Sync status updates can make stale locations look fresh

`OperationalLocationRepository.shouldRefresh()` uses `updatedAtEpochMillis` to decide whether the cached operational location is older than the 30-minute throttle.

However, `markSyncStatus()` updates `updatedAtEpochMillis` whenever sync fails:

```kotlin
val pending = location.copy(
    syncStatus = syncStatus,
    updatedAtEpochMillis = System.currentTimeMillis()
)
```

This creates a stale-location bug:

1. A location was captured a long time ago.
2. It is still pending sync.
3. App launch/resume calls `syncPendingIfAuthenticated()`.
4. Sync fails, for example because the backend endpoint is unavailable or network fails.
5. `markSyncStatus()` updates `updatedAtEpochMillis` to now.
6. `shouldRefresh()` now thinks the old location is fresh.
7. The app skips capturing a fresh GPS location.

This is dangerous because operational location freshness should be based on when the location was actually captured/refreshed, not when a sync attempt happened.

Please fix this by either:

- not modifying `updatedAtEpochMillis` when only changing `syncStatus`, or
- adding separate sync metadata fields such as:
  - `lastSyncAttemptAtEpochMillis`
  - `lastSyncedAtEpochMillis`

The freshness/throttle check should use the actual location capture/update timestamp, not sync attempt time.

### 2. Successful sync also changes location freshness timestamp

In `syncLocation()`, after a successful PATCH, the code also sets:

```kotlin
val synced = location.copy(
    syncStatus = SyncStatus.SYNCED,
    updatedAtEpochMillis = System.currentTimeMillis()
)
```

This has the same problem: syncing an old location successfully should not make the location itself fresh. Sync time and location capture time are different concepts.

Please keep location freshness tied to the actual captured/current location timestamp.

### 3. Missing tests for the new freshness/throttle behavior

The PR deletes the old `AppLaunchLocationUpdaterTest`, but does not add replacement tests for the new operational location logic.

Since this PR introduces important freshness behavior, please add at least minimal tests for:

- no cached location -> should refresh
- cached location younger than 30 minutes -> should not refresh
- cached location older than 30 minutes -> should refresh
- failed pending sync must not update the freshness timestamp
- successful sync must not update the freshness timestamp
- guest mode must not call the authenticated backend sync path

The most important regression test is:

```text
old pending operational location + failed sync attempt should still be considered stale
```

### 4. Backend endpoint dependency should be clarified

This PR calls:

```text
/operational-location/me
```

but the PR notes say that this PR does not modify backend APIs.

If the backend endpoint from the backend operational-location issue is not merged yet, authenticated launch/resume sync will fail and remain pending. That is okay as a temporary dependency, but the PR should clearly state that it depends on the backend endpoint.

Please either:

- mention the dependency explicitly in the PR body, or
- avoid marking authenticated backend sync as fully complete until the backend endpoint exists.

### 5. Consider account-switch/logout behavior for the local cache

The local table stores a single `current` operational location entry. This is probably acceptable for now as a device-level cache, but there is a possible edge case:

1. User A logs in and operational location is stored/pending.
2. User A logs out.
3. User B logs in.
4. The old pending operational location may be synced under User B unless cleared or scoped.

This may be follow-up work, but please either:

- clear operational location on logout/account switch, or
- scope operational location by user/session, or
- explicitly document this as follow-up.

gulcetahtasiz:
Addressed the review points:
- Sync status changes no longer update the operational location freshness timestamp.
- Successful sync also preserves the original location freshness timestamp.
- Added unit tests for throttle/freshness behavior and guest sync gating.
- Cleared operational location cache on logout/account switch.
- Clarified backend dependency in the PR body.

could you re-check @kckagancan ?

kckagancan (approved):
Thanks for the fixes. Everything looks fine now. Merging.

PR #447, opened 2026-05-05
feat: add stale location guard and availability expiry

### LINKED ISSUE #435: [Feature][Backend] Add Stale Location Guard and Availability Session Expiry to Volunteer Matching
## Summary

Volunteer assignment should not use very old operational locations. A volunteer who enabled availability two days ago must not receive an assignment near their old location.

Add location freshness checks and availability session expiry to backend matching.

## Problem

Matching currently sorts candidates by distance and `location_updated_at`, but does not prevent assignment from stale locations.

Example failure:

1. User enables availability in Istanbul.
2. User leaves the app and travels to another city.
3. Two days later, a help request opens near the old Istanbul location.
4. Backend still considers the volunteer available and nearby.
5. Wrong assignment occurs.

## Product Decision

Volunteer availability is a time-limited operational session, not a permanent profile setting.

Suggested defaults:

```text
Operational location refresh throttle: 30 minutes
Max assignable location age: 2 hours
Availability session TTL: 6 hours
```

These should be configurable via env variables.

## Backend Data Changes

Add fields to `volunteers`, or equivalent:

- `available_until`
- `availability_confirmed_at`
- `last_location_accuracy_meters`
- `last_location_source`

Exact naming may follow project conventions.

## Matching Rules

A volunteer is eligible for automatic assignment only if:

- `is_available = TRUE`
- volunteer has no active assignment
- `available_until` is null-safe and not expired
- `last_known_latitude` and `last_known_longitude` exist
- `location_updated_at` exists
- `location_updated_at` is not older than max assignable location age

Sorting should continue to prefer:

1. required expertise / first-aid capability
2. lower distance
3. fresher location
4. deterministic volunteer id tie-breaker

## Availability Session Rules

When volunteer turns availability ON:

- set `availability_confirmed_at = now`
- set `available_until = now + TTL`
- update operational location

When volunteer operational location refreshes while available:

- update location fields
- refresh `availability_confirmed_at`
- extend `available_until = now + TTL`

When availability expires:

- volunteer should be treated as unavailable for matching
- UI/status endpoint should surface that availability expired or was paused
- avoid assigning expired volunteers

## Tasks

- Add DB migration.
- Update availability repository/service.
- Update matching query/candidate filters.
- Add env configuration:
  - `VOLUNTEER_LOCATION_MAX_AGE_MINUTES`
  - `VOLUNTEER_AVAILABILITY_TTL_MINUTES`
- Update `/availability/status` response to include:
  - availableUntil
  - locationUpdatedAt
  - locationFreshness/state if useful
- Add backend tests:
  - fresh nearby volunteer assigned
  - stale volunteer skipped
  - expired available volunteer skipped
  - volunteer with null coordinates skipped
  - first-aid/expertise logic still works
  - existing assignment constraints still work

## Acceptance Criteria

- Volunteer with 2-day-old location is not assigned.
- Volunteer with null location is not assigned to location-based matching.
- Expired availability is not considered assignable.
- Fresh nearby volunteer is still assigned correctly.
- Matching remains deterministic.
- Existing special matching rules for first aid/search and rescue remain intact.

kckagancan:
## Summary

This PR adds stale-location and availability-session guards to volunteer matching.

Volunteer availability is now treated as a time-limited operational session instead of a permanent assignable state.

## Changes

- Added volunteer availability session metadata:
  - `available_until`
  - `availability_confirmed_at`
  - `last_location_accuracy_meters`
  - `last_location_source`
- Added configurable backend defaults for:
  - max assignable location age
  - availability session TTL
- Updated availability ON behavior to refresh session expiry and operational location metadata
- Updated matching eligibility to skip volunteers with:
  - expired availability sessions
  - stale operational locations
  - missing coordinates
  - missing `location_updated_at`
- Updated availability status response with session/freshness fields
- Added backend tests for stale location, expired availability, null location, and existing matching regressions

## Important Behavior

- Volunteers with old operational locations are no longer auto-assigned
- Expired availability sessions are ignored by matching
- Availability OFF still preserves existing location data
- Existing first-aid / SAR matching behavior remains intact

## Testing

- Added/updated backend unit tests
- Added/updated backend integration tests
- Ran relevant backend test commands

Closes #435

### DIFF
--- a/backend/migrations/20260505_120000__add_volunteer_availability_session_metadata.sql
+++ b/backend/migrations/20260505_120000__add_volunteer_availability_session_metadata.sql
@@ -1,0 +1,23 @@
+BEGIN;
+
+ALTER TABLE volunteers
+  ADD COLUMN IF NOT EXISTS available_until TIMESTAMP,
+  ADD COLUMN IF NOT EXISTS availability_confirmed_at TIMESTAMP,
+  ADD COLUMN IF NOT EXISTS last_location_accuracy_meters DOUBLE PRECISION,
+  ADD COLUMN IF NOT EXISTS last_location_source VARCHAR(100);
+
+ALTER TABLE volunteers
+  ADD CONSTRAINT chk_volunteer_last_location_accuracy
+  CHECK (
+    last_location_accuracy_meters IS NULL
+    OR last_location_accuracy_meters >= 0
+  )
+  NOT VALID;
+
+ALTER TABLE volunteers
+  VALIDATE CONSTRAINT chk_volunteer_last_location_accuracy;
+
+CREATE INDEX IF NOT EXISTS idx_volunteers_matching_availability_session
+  ON volunteers (is_available, available_until, location_updated_at);
+
+COMMIT;
--- a/backend/src/config/env.js
+++ b/backend/src/config/env.js
@@ -58,6 +58,10 @@ const env = {
     guestMatchingEnabled: (process.env.HELP_REQUEST_GUEST_MATCHING_ENABLED || 'false').toLowerCase() === 'true',
     guestTokenTtl: process.env.HELP_REQUEST_GUEST_TOKEN_TTL || '2h',
   },
+  volunteerMatching: {
+    locationMaxAgeMinutes: readNumber(process.env.VOLUNTEER_LOCATION_MAX_AGE_MINUTES, 120),
+    availabilityTtlMinutes: readNumber(process.env.VOLUNTEER_AVAILABILITY_TTL_MINUTES, 360),
+  },
 };
 
 module.exports = {
--- a/backend/src/modules/availability/repository.js
+++ b/backend/src/modules/availability/repository.js
@@ -436,24 +436,31 @@ async function updateVolunteerAvailability(
   const ttlMinutes = Number.isFinite(options.availabilityTtlMinutes) && options.availabilityTtlMinutes > 0
     ? Math.floor(options.availabilityTtlMinutes)
     : env.volunteerMatching.availabilityTtlMinutes;
+  const locationTimestamp = typeof options.locationTimestamp === 'string' && options.locationTimestamp.trim() !== ''
+    ? options.locationTimestamp
+    : null;
 
   const sql = `
+    WITH availability_event AS (
+      SELECT COALESCE($8::timestamp, CURRENT_TIMESTAMP) AS occurred_at
+    )
     UPDATE volunteers
     SET is_available = $2,
         last_known_latitude = $3,
         last_known_longitude = $4,
-        location_updated_at = CURRENT_TIMESTAMP,
-        availability_confirmed_at = CURRENT_TIMESTAMP,
-        available_until = CURRENT_TIMESTAMP + ($7::int * INTERVAL '1 minute'),
+        location_updated_at = availability_event.occurred_at,
+        availability_confirmed_at = availability_event.occurred_at,
+        available_until = availability_event.occurred_at + ($7::int * INTERVAL '1 minute'),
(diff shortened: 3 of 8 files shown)


gulcetahtasiz (changes requested):
Thanks for the PR. The overall direction is good: matching now has SQL-level guards for expired sessions, stale locations, null coordinates, and null `location_updated_at`.

I found two correctness issues that should be fixed before merge.

1. Offline sync can make a stale captured location look fresh

`syncAvailability()` receives queued records with `timestamp`, and Android also sends `capturedAt` for the captured device location. But the backend ignores those timestamps when refreshing volunteer location/session metadata:

- `backend/src/modules/availability/service.js:175-180`
- `backend/src/modules/availability/service.js:303-316`
- `backend/src/modules/availability/repository.js:440-447`

`updateVolunteerAvailability()` always writes:

```js
location_updated_at = CURRENT_TIMESTAMP
availability_confirmed_at = CURRENT_TIMESTAMP
available_until = CURRENT_TIMESTAMP + ...
```

This can make an old offline GPS snapshot become fresh at sync time, which defeats the stale-location guard.

Please use the actual location capture/event time for freshness, preferably `capturedAt` when present, otherwise the record `timestamp`, and reject or avoid enabling availability if that captured/event time is already older than `VOLUNTEER_LOCATION_MAX_AGE_MINUTES`.

Please add a regression test for:
- sync availability ON with coordinates but an old `timestamp` / `capturedAt`
- volunteer remains not effectively assignable
- `location_updated_at` is not refreshed to current server time for an old captured location

2. `effectiveIsAvailable` can disagree with matching eligibility when coordinates are missing

Matching now correctly excludes volunteers with missing coordinates:

- `backend/src/modules/availability/repository.js:497-503`
- `backend/src/modules/availability/repository.js:696-702`

But the availability status calculation does not check coordinates:

- `backend/src/modules/availability/service.js:205-231`

So a volunteer row with `is_available = true`, active `available_until`, fresh `location_updated_at`, but null `last_known_latitude/longitude`, can return `effectiveIsAvailable: true` while matching will never assign them.

Since this PR introduces “effective availability”, that field should reflect the same assignability guard as matching. Please include coordinate presence in `effectiveIsAvailable` / freshness status and add a status test for an available volunteer with null coordinates.

Local verification:
- `backend/tests/unit/modules/availability/validators.test.js`
- `backend/tests/unit/modules/availability/service.test.js`

Both passed locally.

gulcetahtasiz:
Thanks for the update. The previous review points look addressed.

I noticed one remaining timestamp edge case: future `capturedAt` / `timestamp` values are accepted and then written directly as `location_updated_at`, `availability_confirmed_at`, and the base for `available_until`.

This means a client with a wrong device clock, or a bad payload, could create a future availability session and appear fresh/assignable longer than intended.

Could we reject timestamps that are meaningfully in the future, or clamp the event time to server `CURRENT_TIMESTAMP` with a small clock-skew allowance? A regression test for future `capturedAt` / `timestamp` on sync availability on would be good. @kckagancan

kckagancan:
Thanks again for catching this. I pushed a fix for the future timestamp edge case.

Availability sync now rejects meaningfully future `timestamp` / `capturedAt` values with `VALIDATION_ERROR`, using a small clock-skew allowance. I also added regression coverage for future `capturedAt` and future `timestamp` cases to ensure they cannot update `location_updated_at`, `availability_confirmed_at`, or extend `available_until`.

Could you re-review when you have time? @gulcetahtasiz

gulcetahtasiz:
Thanks for the update. I re-reviewed the latest commit and the future timestamp issue looks addressed.
The previous review points also still look resolved to me. Approving and merging.

Issue #433, opened 2026-05-04
[Feature][Android] Use Operational Location Snapshots for Guest and Authenticated Request Help


kckagancan:
## Summary

Request Help should use the user's operational/current emergency location, not residential/profile location, as the event location.

The selected/captured operational location must be copied into the help request payload as a snapshot so later movement does not change the request location.

## Problem

The data layer supports request coordinates, but the Request Help form does not consistently preserve or submit coordinates.

Current issues:

- Guest "Use Current Location" can reverse-geocode and autofill fields, but the form state does not store latitude/longitude.
- Map selection can update fields, but selected coordinates are not included in the final submission.
- Authenticated manual Request Help flow mainly pre-fills residential/profile fields.
- Emergency draft is closer to correct behavior, but permission request is missing and profile coordinate fallback should be reconsidered.

## Product Rules

- Request Help location means the emergency/event location.
- It should prefer fresh current operational location.
- It may use manually selected map coordinates.
- It may use manually entered administrative fields if GPS is unavailable.
- It must not silently use residential/profile location as if it were current emergency location.
- Residential/profile location may prefill administrative fields only as a convenience, but the user must be able to override them.

### Update form state

Add coordinate fields to `RequestHelpFormState`:

- latitude
- longitude
- coordinateSource
- coordinateCapturedAt
- coordinateAccuracyMeters, if supported by backend payload

### Use Current Location

- Show `Use Current Location` for both guest and authenticated users.
- If permission is missing, request foreground location permission.
- Capture current device location.
- Reverse-geocode it.
- Store both:
  - coordinate snapshot
  - administrative fields when reverse geocoding succeeds
- If reverse geocoding fails, keep coordinates and ask user to fill administrative fields manually.

### Map picker

- Selecting a point on the map should store selected latitude/longitude in the request form state.
- Reverse-geocode selected point when possible.
- If reverse geocoding fails, keep coordinates and allow manual fields.

### Submission

- `buildSubmission()` must include latitude/longitude/source/capturedAt when present.
- Offline local entity must preserve coordinates.
- Sync operation payload must preserve coordinates.
- Updating a draft must preserve coordinates unless the user explicitly changes location.

### Authenticated flow

- Authenticated users should not be limited to residential profile location.
- Profile location can prefill fields, but the request should clearly ask for emergency location.
- Button/subtitle copy should change from "Use the same location structure as your profile" to something like:
  - "Where do you need help right now?"

### Guest flow

[...]

PR #453, opened 2026-05-05
feat: use operational location snapshots for request help

gulcetahtasiz:
## Summary

Implements Android-side operational/current location snapshots for Request Help.

Request Help now treats location as the emergency/event location, not residential/profile location. Current location and map-selected coordinates are copied into the request payload as a snapshot so later device movement does not change the request location.

## Changes

- Added coordinate snapshot fields to Request Help form state
- Stores current GPS coordinates when Use Current Location succeeds
- Requests foreground location permission from Request Help when needed
- Reverse-geocodes current/map-selected coordinates when possible
- Preserves coordinates even when reverse geocoding fails
- Stores map-selected coordinates in form state and submission payload
- Includes coordinate source, capturedAt, and accuracyMeters in local/offline request data
- Preserves coordinates in draft edit/update flow unless changed
- Updates Request Help UI copy to clarify emergency/event location
- Removes profile shared-coordinate fallback from emergency draft creation
- Adds Room migration for request coordinate accuracy metadata

## Testing

- Ran targeted Request Help Android unit tests
- Ran Android debug Kotlin compile

Closes [#433](https://github.com/bounswe/bounswe2026group6/issues/433)

### DIFF
--- a/android/app/src/main/java/com/neph/core/database/NephDatabase.kt
+++ b/android/app/src/main/java/com/neph/core/database/NephDatabase.kt
@@ -17,7 +17,7 @@ import com.neph.BuildConfig
         SyncOperationEntity::class,
         SyncMetadataEntity::class
     ],
-    version = 5,
+    version = 6,
     exportSchema = false
 )
 abstract class NephDatabase : RoomDatabase() {
@@ -76,6 +76,11 @@ object NephDatabaseProvider {
             )
         }
     }
+    private val Migration5To6 = object : Migration(5, 6) {
+        override fun migrate(database: SupportSQLiteDatabase) {
+            database.execSQL("ALTER TABLE help_requests ADD COLUMN coordinateAccuracyMeters REAL")
+        }
+    }
 
     fun initialize(context: Context) {
         getInstance(context)
@@ -87,7 +92,7 @@ object NephDatabaseProvider {
                 context.applicationContext,
                 NephDatabase::class.java,
                 DatabaseName
-            ).addMigrations(Migration1To2, Migration2To3, Migration3To4, Migration4To5)
+            ).addMigrations(Migration1To2, Migration2To3, Migration3To4, Migration4To5, Migration5To6)
                 .build()
                 .also { instance = it }
         }
--- a/android/app/src/main/java/com/neph/core/database/OfflineEntities.kt
+++ b/android/app/src/main/java/com/neph/core/database/OfflineEntities.kt
@@ -36,6 +36,7 @@ data class HelpRequestEntity(
     val longitude: Double? = null,
     val coordinateSource: String? = null,
     val coordinateCapturedAt: String? = null,
+    val coordinateAccuracyMeters: Double? = null,
     val contactFullName: String,
     val contactPhone: String,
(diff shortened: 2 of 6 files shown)


delibas-rojhat (changes requested):
Thanks for the implementation.

I noticed two location consistency issues that should be fixed before merge:

1. If the user captures/selects coordinates using **Use Current Location** or the map picker, and then manually changes the location fields afterward, the previously captured coordinate snapshot is still kept and submitted.

   This can produce an inconsistent request where the visible administrative location points to one place, but the submitted latitude/longitude still points to an older GPS/map location.

   A possible fix would be to clear the coordinate snapshot whenever the user manually changes country/city/district/neighborhood/address fields, or explicitly mark the coordinates as stale and ask the user to re-confirm/update them.

2. In the quick emergency draft flow, if current GPS coordinates are captured but reverse geocoding fails or returns incomplete administrative fields, the draft can still fall back to residential/profile location fields.

   This can mix current emergency coordinates with home/profile administrative fields, which conflicts with the requirement that Request Help must not silently treat residential/profile location as the emergency/event location.

   A safer behavior would be to avoid using profile administrative fields for the emergency draft when reverse geocoding fails, and instead open the Request Help form with the captured coordinates preserved so the user can manually confirm the event location fields.

kckagancan (changes requested):
Thanks for the follow-up fix. The quick emergency draft fallback is much safer now, and the UI now clears the coordinate snapshot when manual location fields change.

I still found one blocking consistency issue before approval:

When editing an existing draft that already has coordinates, `updateManualLocation(...)` clears the coordinate snapshot from the form state, but `RequestHelpRepository.updateHelpRequest(...)` later calls `submission.withPreservedCoordinates(existing)`. Since the new submission has no latitude/longitude, the repository copies the old draft coordinates back into the update payload.

So the stale coordinate bug can still happen in draft edit/update flow:

1. Create or open a draft with GPS/map coordinates.
2. Manually change country/city/district/neighborhood/address.
3. UI clears coordinates correctly.
4. Submit the draft update.
5. Repository re-adds the previous coordinates from the existing entity.

This violates the “preserve coordinates unless changed” requirement. We need an explicit way to distinguish “coordinates missing because the user did not touch location” from “coordinates intentionally cleared because the user changed the event location.”

Possible fix:
- Track a `locationWasManuallyChanged` / `coordinateSnapshotClearedByLocationEdit` flag in form state, and pass it to `updateHelpRequest`.
- Only call `withPreservedCoordinates(existing)` when location was not changed.
- Add tests for:
  - draft update preserves coordinates when only non-location fields change
  - draft update clears/does not preserve coordinates when location fields change

One more related edge case: in authenticated Request Help, profile location fields are prefilled. If the user captures current location or selects map coordinates but reverse geocoding fails/returns incomplete data, the new coordinates can remain paired with old profile administrative fields. Since the issue says residential/profile location must not be silently treated as the emergency/event location, GPS/map coordinate flows should clear or replace the admin fields when reverse geocoding cannot safely resolve them.

kckagancan (approved):
Looks good to me now. Merging.

PR #454, opened 2026-05-05
fix: parse country-specific Android location tree

### LINKED ISSUE #450: [Bug] Fix Android location tree parsing in Edit Profile
## Description

The Android Edit Profile “Residential Location” dropdown is parsing the backend location tree incorrectly.

Observed behavior:
- In the Country dropdown, the only real option shown is `cities`.
- City/district/neighborhood selection is unusable because Android treats `cities` as a country.
- This is visible in Edit Profile → Residential Location.

Root cause:
- Backend `GET /location/tree?countryCode=TR` returns a country-specific tree:
  `{ tree: { label: "Turkey", cities: { ... } } }`
- Android `LocationTreeRepository.parseLocationTreeResponse(...)` expects `tree` to be keyed by country code:
  `{ tree: { TR: { label: "Turkey", cities: { ... } } } }`
- Because of that mismatch, Android iterates `label` and `cities` as if they were country keys, and `cities` becomes a bogus country option.

Expected behavior:
- Country dropdown should show `Turkey`, not `cities`.
- City/district/neighborhood dropdowns should populate correctly.
- If the backend response is country-specific, Android should wrap it under the requested country code before converting it to `LocationData`.
- Add a unit test for the actual backend response shape returned by `/location/tree?countryCode=TR`.

**Also, I am not sure if it is under construction, or listed in some other issues, but, it seems to me that the following failure seems to persist.**

Image

kckagancan:
## Summary

Fixes Android location tree parsing for the Edit Profile residential location dropdown.

The backend `/location/tree?countryCode=TR` endpoint can return a country-specific tree in this shape:

`{ "tree": { "label": "Turkey", "cities": { ... } } }`

Android previously expected the tree to always be keyed by country code:

`{ "tree": { "TR": { "label": "Turkey", "cities": { ... } } } }`

Because of this mismatch, the parser treated `cities` as a country option, causing the Country dropdown to show `cities` instead of `Turkey`.

## Changes

- Updated Android location tree parsing to support country-specific backend responses.
- Preserved support for the existing country-keyed tree shape.
- Ensured country-specific responses are normalized under the requested country code.
- Added focused unit tests for:
  - country-specific `/location/tree?countryCode=TR` response shape
  - existing country-keyed response shape
  - empty tree fallback behavior

## Expected Behavior

- Edit Profile → Residential Location shows `Turkey` in the Country dropdown.
- City, district, and neighborhood dropdowns populate correctly.
- Existing fallback/static location data behavior remains unchanged.

## Testing

- Android unit tests were run for the affected parser behavior.

Closes #450

### DIFF
--- a/android/app/src/main/java/com/neph/features/profile/data/LocationTreeRepository.kt
+++ b/android/app/src/main/java/com/neph/features/profile/data/LocationTreeRepository.kt
@@ -40,7 +40,10 @@ object LocationTreeRepository {
             path = "/location/tree?countryCode=$normalizedCountryCode"
         )
 
-        val parsed = parseLocationTreeResponse(response)
+        val parsed = parseLocationTreeResponse(
+            response = response,
+            requestedCountryCode = normalizedCountryCode
+        )
         if (parsed.isEmpty()) {
             throw ApiException(
                 message = "Location options are unavailable right now.",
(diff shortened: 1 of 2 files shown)


gulcetahtasiz (approved):
The fix looks good to me. The Android parser now correctly handles the country-specific `/location/tree?countryCode=...` response shape while preserving the previous country-keyed tree behavior. The added tests cover both response formats plus empty/malformed trees.
Approving.

PR #456, opened 2026-05-05
feat: cache nearby visible users for offline access

### LINKED ISSUE #387: [Feature] Local Caching of Nearby Visible Users
## Description
Implement local caching of nearby visible users on Android so the app can provide faster access and partial functionality during weak or unavailable connectivity.

Cached data should only include users that the current user is allowed to see based on privacy and trusted circle rules. The app should clearly indicate when cached data may be stale.

## Scope
Android, Backend integration

## Related Issues
- Related to #380
- Related to #382

## Tasks
- [ ] Define which nearby user fields are safe to cache.
- [ ] Reuse or add backend endpoint for retrieving nearby visible users.
- [ ] Implement Android local persistence for nearby user cache.
- [ ] Store only minimal required data.
- [ ] Store cache timestamps.
- [ ] Show cached data when network requests fail.
- [ ] Clearly label stale cached data.
- [ ] Add cache expiration or cleanup behavior.
- [ ] Ensure privacy rules are applied before caching.
- [ ] Avoid caching exact location unless explicitly allowed.
- [ ] Add tests for cache read/write and stale data behavior where applicable.

## Suggested Cached Fields
- User ID
- Display name
- Safety status
- Last updated time
- Approximate/coarse location if allowed
- Visibility scope metadata if needed

## Acceptance Criteria
- [ ] Android caches nearby visible users after successful sync.
- [ ] Cached data is available when offline or network requests fail.
- [ ] Stale data is clearly shown as stale.
- [ ] Cache expiration or refresh policy is defined.
- [ ] The app does not cache users who should not be visible to the current user.
- [ ] Exact location is not cached unless explicitly allowed.
- [ ] Cache behavior does not break normal online sync.

## Notes
This should be privacy-first and Android-first because offline behavior is most important on mobile during disasters.

### LINKED ISSUE #438: [FEATURE] Keep Circle and Nearby Features Residential-Location Based
## Summary

Ensure Family/Circle Safety Check-in (#382) and Nearby Visible Users cache (#387) use the correct location semantics.

These features should not accidentally use current operational GPS as the default "nearby/community" basis.

## Related Issues

- Related to #382
- Related to #387
- Related to #380
- Related to #384

### Family / Circle Safety Check-in

Circle membership and visibility should be relationship/trusted-circle based.

Location may appear only as:

- optional safety status/check-in location snapshot
- shared only with explicit consent
- controlled by privacy/trusted circle rules

Circle membership/discovery should not depend on current operational GPS unless a future issue explicitly defines such behavior.

### Nearby Visible Users

Nearby visible users / nearby neighbors should be based on residential/profile area, not current operational GPS.

Suggested basis:

- country
- city
- district
- neighborhood
- privacy/trusted-circle visibility rules

Operational/current location should not be cached or used for nearby-neighbor discovery by default.

## Rules for #382

- Use trusted circle relationship as the main visibility mechanism.
- Show latest safety status.
- Show latest check-in timestamp.
- Show location only if:
  - user explicitly shared location with that safety status
  - viewer is allowed by privacy/trusted circle rules
- Do not use volunteer operational location for circle member location.
- Do not overwrite residential/profile location.

## Rules for #387

- Cache only fields the user is allowed to see.
- Use residential/profile area for "nearby".
- Cache coarse area only when allowed.
- Do not cache exact operational GPS unless a future explicit feature requires it.
- Clearly label stale cached data.
- Include cache timestamps and expiration policy.
- Respect privacy before writing to local cache.

## Tasks

- Review #382 implementation when available.
- Review #387 implementation when available.
- Add missing guardrails if current implementation uses operational/current GPS incorrectly.
- Update UI copy to avoid ambiguous "nearby" meaning.
- Add tests where possible:
  - user outside current GPS area still sees residential nearby neighbors
  - hidden/private users are not cached
  - safety status location is visible only with consent/privacy
  - exact location is not cached by default

## Acceptance Criteria

- #382 does not use operational location for circle membership/discovery.
- #387 does not use operational location as default nearby-neighbor basis.
- Residential/profile area drives nearby community features.
- Safety status location is treated as optional check-in snapshot only.
- Cached nearby visible users remain privacy-first and coarse.

delibas-rojhat:
**Closes**

- Closes #387
- Closes #438

**Summary**

Implements Android local caching for nearby visible users with privacy-first backend filtering.

**Changes**

- Adds `nearby=true` support to the visible safety statuses endpoint.
- Filters nearby users by shared residential/profile neighborhood context while preserving existing visibility/trusted-circle privacy rules.
- Does not expose `city`, `district`, or `neighborhood` fields in visible safety status responses.
- Includes nearby users based on profile-area visibility; exact/coarse location is returned only when location sharing allows it.
- Adds Android Room cache for nearby visible users.
- Scopes cached nearby users by the current viewer/cache owner to avoid cross-account cache leaks.
- Stores minimal cached fields only.
- Stores cache timestamps and expiration metadata.
- Shows cached users only for network/non-auth failures.
- Rethrows `401` auth errors so the screen can trigger logout/login handling.
- Labels offline/stale cached data in the Nearby Users screen.
- Avoids caching exact location unless the backend says location is visible.
- Adds Android cache tests and backend integration coverage.
- Rebases Android Room migration onto current development as DB version `9` with `Migration8To9`.

### DIFF
--- a/android/app/src/main/java/com/neph/core/database/NephDatabase.kt
+++ b/android/app/src/main/java/com/neph/core/database/NephDatabase.kt
@@ -15,10 +15,11 @@ import com.neph.BuildConfig
         OperationalLocationEntity::class,
         SafetyStatusEntity::class,
         AssignedRequestEntity::class,
+        NearbyVisibleUserEntity::class,
         SyncOperationEntity::class,
         SyncMetadataEntity::class
     ],
-    version = 8,
+    version = 9,
     exportSchema = false
 )
 abstract class NephDatabase : RoomDatabase() {
@@ -27,6 +28,7 @@ abstract class NephDatabase : RoomDatabase() {
     abstract fun operationalLocationDao(): OperationalLocationDao
     abstract fun safetyStatusDao(): SafetyStatusDao
     abstract fun assignedRequestDao(): AssignedRequestDao
+    abstract fun nearbyVisibleUserDao(): NearbyVisibleUserDao
     abstract fun syncOperationDao(): SyncOperationDao
     abstract fun syncMetadataDao(): SyncMetadataDao
 }
@@ -119,6 +121,31 @@ object NephDatabaseProvider {
             database.execSQL("ALTER TABLE availability_state ADD COLUMN pauseReason TEXT NOT NULL DEFAULT 'NONE'")
         }
     }
+    private val Migration8To9 = object : Migration(8, 9) {
+        override fun migrate(database: SupportSQLiteDatabase) {
+            database.execSQL(
+                """
+                CREATE TABLE IF NOT EXISTS nearby_visible_users (
+                    cacheOwnerUserId TEXT NOT NULL,
+                    userId TEXT NOT NULL,
+                    displayName TEXT,
+                    safetyStatus TEXT NOT NULL,
+                    statusUpdatedAt TEXT,
+                    latitude REAL,
+                    longitude REAL,
+                    locationCapturedAt TEXT,
+                    visibilityScope TEXT,
+                    fetchedAtEpochMillis INTEGER NOT NULL,
+                    expiresAtEpochMillis INTEGER NOT NULL,
+                    PRIMARY KEY(cacheOwnerUserId, userId)
+                )
+                """.trimIndent()
+            )
+            database.execSQL("CREATE INDEX IF NOT EXISTS index_nearby_visible_users_cacheOwnerUserId ON nearby_visible_users (cacheOwnerUserId)")
+            database.execSQL("CREATE INDEX IF NOT EXISTS index_nearby_visible_users_safetyStatus ON nearby_visible_users (safetyStatus)")
+            database.execSQL("CREATE INDEX IF NOT EXISTS index_nearby_visible_users_fetchedAtEpochMillis ON nearby_visible_users (fetchedAtEpochMillis)")
+        }
+    }
 
     fun initialize(context: Context) {
         getInstance(context)
@@ -137,7 +164,8 @@ object NephDatabaseProvider {
                 Migration4To5,
                 Migration5To6,
                 Migration6To7,
-                Migration7To8
+                Migration7To8,
+                Migration8To9
             )
                 .build()
                 .also { instance = it }
--- a/android/app/src/main/java/com/neph/core/database/OfflineDaos.kt
+++ b/android/app/src/main/java/com/neph/core/database/OfflineDaos.kt
@@ -121,6 +121,48 @@ interface AssignedRequestDao {
 }
 
 @Dao
+interface NearbyVisibleUserDao {
+    @Query(
+        """
+        SELECT * FROM nearby_visible_users
+        WHERE cacheOwnerUserId = :cacheOwnerUserId
+        ORDER BY fetchedAtEpochMillis DESC, displayName ASC, userId ASC
+        """
+    )
+    fun observeByCacheOwner(cacheOwnerUserId: String): Flow<List<NearbyVisibleUserEntity>>
+
+    @Query(
+        """
+        SELECT * FROM nearby_visible_users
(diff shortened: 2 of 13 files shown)


kckagancan (changes requested):
Thanks, this is a solid start and the main direction matches #387/#438: nearby users are based on residential/profile area rather than operational GPS, Android caches minimal nearby user fields, stale cache is labelled, and CI is green.

I found a few privacy/behavior issues that I think should be fixed before approval:

1. The Android nearby users cache is not scoped to the current viewer/user.
   `NearbyVisibleUserEntity` is keyed only by `userId`, and offline fallback reads all cached rows. This can leak User A’s cached nearby users to User B after logout/login or token/account switch if the next refresh fails. The cache should include a viewer/cache owner id and all reads/writes/clears should be scoped to the current authenticated user, or the cache must be cleared reliably on logout/user switch.

2. `refreshNearbyVisibleUsers` catches all exceptions except cancellation and falls back to cache. This also swallows auth errors such as 401. The screen has 401 handling, but it will not run if the repository converts the 401 into cached data. Please rethrow `ApiException` with status 401 and only use cache fallback for real network/non-auth failures.

3. Backend `mapVisibleSafetyStatus` now returns `city`, `district`, and `neighborhood` unconditionally. This changes the normal `/safety-status/visible` response and may leak residential location fields outside the existing location visibility logic. The SQL can use residential fields internally for filtering, but the response should not expose them unless explicitly allowed. Since Android does not need these fields for caching, I would remove them from the response or guard them behind the same privacy rules.

4. Please double-check the product semantics around `nearby=true`: the current SQL requires `location_sharing_enabled = true` and `location_visibility != PRIVATE` for non-admin nearby results. If nearby discovery is supposed to be residential/profile-area based, users should probably be included based on profile/trusted-circle visibility and same residential area, while exact/coarse location should only be cached/displayed when location sharing allows it. If the intended behavior is to fully exclude users with private location, the issue/PR should state that explicitly.

5. Migration note: this PR bumps Android Room DB from 5 to 6 with `Migration5To6`. If another open Android DB migration PR is merged first, this will need a rebase and likely a `Migration6To7` or combined migration to avoid version conflicts.

kckagancan (changes requested):
Thanks, the previous review items are mostly addressed now. The cache is scoped by `cacheOwnerUserId`, 401 auth errors are rethrown instead of falling back to cache, residential fields are no longer exposed in visible safety status responses, nearby discovery is now based on profile/residential area, and the Room migration is correctly rebased to version 9.

1. Please update the Nearby Users UI copy to make the #438 product semantics explicit.

Right now the screen says:

> Only people visible through privacy or trusted circle rules are cached.

This explains privacy, but it does not explain that “nearby” is based on the user’s residential/profile area rather than current GPS. #438 explicitly asks us to avoid ambiguous nearby wording.

Suggested copy:

> Based on your residential/profile area, not your current GPS. Only users visible through privacy or trusted circle rules are cached.

2. Please add/clarify a backend privacy guardrail test for safety-status location visibility.

`mapVisibleSafetyStatus` returns location when:

```js
share_location_consent &&
location_sharing_enabled &&
location_visibility !== 'PRIVATE'
```

That means EMERGENCY_ONLY location can be returned if the user is otherwise visible. Since #438 says safety-status location should only appear when consent and privacy/trusted-circle rules allow it, please add a focused test for this edge case:

target user has profile_visibility = PUBLIC
target user has location_visibility = EMERGENCY_ONLY
target shares safety-status location
viewer is in the same residential area but is not in the target’s trusted circle

Expected behavior should be either:

location: null, if EMERGENCY_ONLY means trusted/emergency-only visibility, or
an explicit test/comment documenting that this is intentionally visible to nearby safety-status viewers.

Everything else looks aligned with #387/#438, and the earlier cache-owner/auth/migration concerns are fixed.

kckagancan (approved):
Thanks for the fixes. Great work. Merging.

PR #461, opened 2026-05-06
fix: implement fallback content mechanism during API failures

### LINKED ISSUE #388: [Bug] Replace weak API error states with useful fallback states
Several pages show:

  > “Could not reach the server. Please check your connection and try again.”

  This is technically correct, but bad for demos and emergency UX.

  Improve with:

  - clear retry button,
  - explanation of what failed,
  - fallback/cached/demo content,
  - “Last updated” info,
  - skeleton/loading states.

  For example, /news should still show seeded/demo announcements if available, and /
  gathering-areas should show fallback areas or guidance when provider/backend fails.

AlperKartkaya:
- Improves public outage UX for news, home previews, and gathering areas by replacing hard server-error states with skeleton loading, retry actions, last-updated context, and cached/demo fallback content. Gathering areas now stays usable during backend/provider failures with cached results or demo guidance instead of blank failure states.

Resolves #388

### DIFF
--- a/web/e2e/gathering-areas.spec.js
+++ b/web/e2e/gathering-areas.spec.js
@@ -115,7 +115,7 @@ test('keeps map/list selection stable when features share same id but different 
   await expect(page.locator('.gathering-areas-selected-meta').first()).toContainText('Shelter');
 });
 
-test('shows empty and error states for gathering areas retrieval', async ({ page }) => {
+test('shows empty and fallback states for gathering areas retrieval', async ({ page }) => {
   await mockGeolocation(page);
 
   let requestCount = 0;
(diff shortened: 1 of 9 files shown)


kckagancan (changes requested):
Requesting small changes before approval.

Overall this PR is in the right direction and covers most of #388: `/news`, home news preview, and `/gathering-areas` now have skeletons, retry actions, last-updated context, and cached/demo fallback content.

The main issue is in gathering areas fallback handling.

### 1. `source: "fallback"` with non-empty features is treated like live success

In `GatheringAreasPage`, `response.source === "fallback"` is only handled when `mapped.length === 0`.

If the backend returns `source: "fallback"` with non-empty fallback features, the UI clears `dataNotice` and renders it like a normal live result. That hides the fact that the live provider/backend fallback is being used.

Please handle `response.source === "fallback"` regardless of whether mapped features are empty:

- if backend fallback features exist, render them but show a warning/notice
- if backend fallback features are empty, use the local demo fallback features
- set `fetchState` to `fallback`
- show retry action and last-updated/fallback guidance

This is important because #388 explicitly asks for useful fallback states, not silent fallback-as-live behavior.

### 2. Home preview could reuse cached announcements before demo fallback

`/news` uses `live -> cached -> demo`, but Home preview currently uses `live -> demo`.

This is not as critical as the gathering areas issue, but since the PR scope includes home previews, it would be more consistent to use cached announcements first when available.

### 3. Please add one focused regression test

A good test would be:

- gathering areas API returns `source: "fallback"` with non-empty features
- UI shows fallback warning/guidance
- UI still renders the fallback features
- retry action is visible

Everything else looks aligned with #388.

AlperKartkaya:
Thanks for the review. 

  #### What's Fixed:

  1. Gathering areas source: "fallback"
      - Now handled regardless of feature count.
      - Non-empty backend fallback features render with a visible warning/notice.
      - Empty backend fallback still falls back to local demo areas.
      - fetchState is set to "fallback".
      - Retry + last-updated/fallback guidance are shown.
  2. Home news preview
      - Now uses live -> cached -> demo, matching /news.
      - Shared announcement cache helpers moved into web/src/lib/news.ts.
  3. Regression test
      - Added focused Playwright test for:
          - backend returns source: "fallback" with non-empty features
          - warning/guidance is visible
          - fallback feature renders
          - retry actions are visible

kckagancan (approved):

PR #462, opened 2026-05-06
feat: implement user account soft deletion and request state cleanup

### LINKED ISSUE #411: [Feature] No account deletion endpoint or UI
Requirement: 1.1.6.4

> 1.1.6.4 Users shall be able to delete their account and associated personal data.

The system has no mechanism for users to delete their account. The backend has is_deleted flag checks everywhere but no route/service function to actually perform a soft-delete. This is both a requirements gap and a potential GDPR/privacy concern.

Acceptance Criteria:

- DELETE /api/auth/me endpoint performs a soft-delete (sets is_deleted = TRUE, nullifies PII)
- Web profile page includes a "Delete Account" action with confirmation modal
- Android settings screen offers the same action
- All active assignments/availability are cancelled before deletion

AlperKartkaya:
- This PR adds account soft-delete via DELETE /api/auth/me, anonymizing personal data and cancelling active help requests, assignments, and volunteer availability. Adds web and Android Delete Account confirmation flows with backend/unit/integration coverage.

Resolves #411

### DIFF
--- a/android/app/src/main/java/com/neph/core/database/OfflineDaos.kt
+++ b/android/app/src/main/java/com/neph/core/database/OfflineDaos.kt
@@ -58,6 +58,9 @@ interface HelpRequestDao {
 
     @Query("DELETE FROM help_requests WHERE localId = :localId")
     suspend fun deleteByLocalId(localId: String)
+
+    @Query("DELETE FROM help_requests WHERE ownerType = :ownerType")
+    suspend fun deleteByOwner(ownerType: String)
 }
 
 @Dao
@@ -70,6 +73,9 @@ interface AvailabilityDao {
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun upsert(entity: AvailabilityEntity)
+
+    @Query("DELETE FROM availability_state")
+    suspend fun clear()
 }
 
 @Dao
@@ -202,6 +208,22 @@ interface SyncOperationDao {
 
     @Query("DELETE FROM sync_operations WHERE status = 'SYNCED'")
     suspend fun deleteSynced()
+
+    @Query("DELETE FROM sync_operations WHERE entityType = :entityType")
+    suspend fun deleteByEntityType(entityType: String)
+
+    @Query(
+        """
+        DELETE FROM sync_operations
+        WHERE entityType = :entityType
+          AND entityId IN (
+              SELECT localId
+              FROM help_requests
+              WHERE ownerType = :ownerType
+          )
+        """
+    )
+    suspend fun deleteHelpRequestOperationsForOwner(entityType: String, ownerType: String)
 }
 
 @Dao
--- a/android/app/src/main/java/com/neph/features/assignedrequest/data/AssignedRequestRepository.kt
+++ b/android/app/src/main/java/com/neph/features/assignedrequest/data/AssignedRequestRepository.kt
@@ -121,6 +121,11 @@ object AssignedRequestRepository {
         return pending.toUiModel()
     }
 
+    suspend fun clearLocalCache() {
+        database.assignedRequestDao().clearAll()
+        database.syncOperationDao().deleteByEntityType(SyncEntityType.ASSIGNED_REQUEST)
+    }
+
     suspend fun fetchAssignmentRoute(token: String, assignmentId: String): AssignmentRouteUiModel? {
         val response = JsonHttpClient.request(
             path = "/assignments/$assignmentId/route",
--- a/android/app/src/main/java/com/neph/features/auth/data/AuthRepository.kt
+++ b/android/app/src/main/java/com/neph/features/auth/data/AuthRepository.kt
@@ -4,13 +4,16 @@ import android.util.Log
 import com.google.firebase.messaging.FirebaseMessaging
 import com.neph.core.NephAppContext
 import com.neph.core.network.ApiException
-import com.neph.core.sync.OfflineSyncScheduler
 import com.neph.core.network.JsonHttpClient
+import com.neph.core.sync.OfflineSyncScheduler
+import com.neph.features.assignedrequest.data.AssignedRequestRepository
+import com.neph.features.availability.data.AvailabilityRepository
 import com.neph.features.notifications.data.PushTokenSync
 import com.neph.features.notifications.data.NotificationsRepository
 import com.neph.features.operationallocation.data.OperationalLocationRepository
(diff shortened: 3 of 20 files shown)


kckagancan (changes requested):
Requesting changes before approval. The overall direction is good and the PR covers the backend endpoint plus web/Android delete account flows, but I found two backend blockers.

### 1. Migration does not match the columns nullified by soft delete

`softDeleteUserAccount()` sets `help_requests.contact_full_name = NULL` and `request_locations.neighborhood = NULL`, but the migration only drops NOT NULL from `help_requests.contact_phone` and `request_locations.country/city/district`.

Please also make these nullable:

- `help_requests.contact_full_name`
- `request_locations.neighborhood`

Otherwise `DELETE /api/auth/me` can fail with a not-null constraint violation when anonymizing existing help requests/request locations.

Please also double-check every column/table touched by `softDeleteUserAccount()` against the actual migration chain.

### 2. Deleting one volunteer can remove other volunteers' assignments on the same request

The current assignment cleanup deletes assignments where `request_id = ANY(affectedRequestIds)` or the assignment belongs to the deleting user's volunteer record.

Because `affectedRequestIds` includes requests assigned to the deleting volunteer, this can delete all active assignments on those requests, including assignments belonging to other volunteers.

Expected behavior:

- If the deleted user owns the help request, cancel the request and cancel its active assignments.
- If the deleted user is only a volunteer on someone else's request, cancel only that user's volunteer assignment.
- Preserve other volunteers' active assignments on the same request.
- Resync request status based on remaining active assignments.

Please add a regression test with a request that has two active volunteer assignments: deleting one volunteer should not cancel/delete the other volunteer's assignment, and the request should remain assigned if another active assignment remains.

### Additional recommendation

Assignments should probably be marked `is_cancelled = TRUE` rather than physically deleted, since the domain already uses `is_cancelled` to represent cancelled assignments and preserve history.

Everything else looks generally aligned with #411: route/service wiring, soft-delete intent, web profile modal, Android settings flow, local auth cleanup, and old-token rejection are in the right direction.

AlperKartkaya:
#### Thanks for the review.
 ### Applied only necessary backend changes

  ### 1. Migration blocker check

  I did not change the migration, because the requested nullable columns are already covered in the
  actual migration chain:

  - help_requests.contact_full_name is made nullable in:
      - backend/migrations/20260429_100000__make_help_request_fields_optional.sql
  - request_locations.neighborhood is made nullable in:
      - backend/migrations/20260429_100000__make_help_request_fields_optional.sql

  The current soft-delete migration still only needs the columns newly required by account deletion:

  - user_profiles.first_name
  - user_profiles.last_name
  - help_requests.contact_phone
  - request_locations.country/city/district

  I also double-checked the other nullified columns against schema/migrations; no additional NOT NULL
  blockers were found.

  ### 2. Fixed assignment cleanup blocker

  Updated softDeleteUserAccount() so it no longer deletes all assignments on a request when the deleted
  user is only one volunteer.

  New behavior:

  - If deleted user owns an open help request:
      - request is cancelled
      - active assignments for that owned request are marked is_cancelled = TRUE
  - If deleted user is only a volunteer on someone else’s request:
      - only that volunteer’s assignment is marked is_cancelled = TRUE
      - other volunteers’ active assignments are preserved
      - request status is resynced from remaining active assignments

  Also changed assignment cleanup from physical DELETE to:

  UPDATE assignments SET is_cancelled = TRUE

  so history is preserved.

  ### 3. Added regression coverage

  Updated the auth integration test to cover:

  - request with two active volunteer assignments
  - deleting one volunteer cancels only that volunteer’s assignment
  - the other volunteer assignment remains active
  - request remains ASSIGNED

kckagancan (changes requested):
Thanks, the previous backend blockers are mostly addressed now.

The migration concern is resolved: `help_requests.contact_full_name` and `request_locations.neighborhood` are already made nullable by `20260429_100000__make_help_request_fields_optional.sql`, so the new migration does not need to repeat them.

The assignment cleanup blocker is also mostly fixed: assignments are now marked `is_cancelled = TRUE` instead of being physically deleted, and the regression test correctly verifies that deleting one volunteer preserves another active volunteer assignment on the same request.

I still want to request two changes before approval.

### 1. Android account deletion should clear local sensitive/offline caches

`AuthRepository.deleteAccount()` calls `clearLocalAuthState()` after `DELETE /auth/me`, but that currently clears only auth/profile/operational-location/safety-status local state.

For account deletion, we should also clear local offline data that may contain the deleted user’s personal data, especially cached authenticated help requests, assigned request cache, and related pending sync/offline state where applicable.

This matters because #411 is about deleting the account and associated personal data, and Android Room cache may still contain help request contact info, descriptions, locations, assigned request data, etc. Also, authenticated help request cache appears to be scoped by `ownerType = AUTHENTICATED`, not by user id, so stale cached data could leak across account switches on the same device.

Please add local cleanup for account deletion success, separate from normal logout behavior if needed.

### 2. Request status resync still has an `IN_PROGRESS` edge case

The updated assignment cleanup resyncs assigned requests, but the current CASE preserves `IN_PROGRESS` unconditionally:

```sql
WHEN hr.status = 'IN_PROGRESS' THEN 'IN_PROGRESS'
```

If the deleted user is the only active volunteer on an IN_PROGRESS request, their assignment becomes cancelled, no active assignment remains, but the request can still stay IN_PROGRESS.

Expected behavior: preserve IN_PROGRESS only if at least one active assignment remains; otherwise move the request back to PENDING.

Please add a regression test for:

request is IN_PROGRESS
deleted user is the only active volunteer assignment
after account deletion, that assignment is cancelled
request status becomes PENDING

Everything else looks aligned with #411.

AlperKartkaya:
Thanks for the review, here is what's changed: 

  - Android account deletion now additionally clears deletion-sensitive local/offline state:
      - authenticated help request Room cache
      - related authenticated help-request sync operations
      - assigned request cache + pending assignment sync ops
      - availability cache + pending availability sync ops
      - operational location cache
      - safety status cache
  - Normal logout behavior remains separate.
  - Backend request status resync now preserves IN_PROGRESS only when at least one active assignment
    remains.
  - Added regression coverage for account deletion where:
      - request is IN_PROGRESS
      - deleted user is the only active volunteer
      - assignment is cancelled
      - request becomes PENDING
  - Updated availability cancellation regression to expect PENDING when no active assignments remain.

kckagancan (approved):

PR #466, opened 2026-05-06
 Release v0.2.0

AlperKartkaya:
## Summary

  Merges `development` into `main` for the `v0.2.0` release.

  This release includes major emergency response, location, safety, admin, notification, news, Android,
  and stability improvements completed after `v0.1.0`.

  ## Highlights

  ### Emergency response and help requests

  - Added advanced help-request matching with multi-responder support
  - Added help-request current/history views with urgency, priority, and lifecycle metadata
  - Added help-request update support and offline Android draft creation
  - Added public active help-request visibility API
  - Added web and Android crisis/help-request map views

  ### Location and gathering areas

  - Added gathering areas on web and Android with map-based nearby results
  - Added cached/demo fallback handling for provider/backend failures
  - Added reusable map components and in-app map picker flows
  - Added operational location snapshots and freshness checks
  - Improved location tree parsing and profile coordinate persistence

  ### Safety and privacy

  - Added safety status tracking with backend persistence
  - Added trusted safety circles and circle owner/member management
  - Added offline safety check-in support
  - Improved location-sharing consent and permission handling
  - Fixed safety status cleanup on account switch

  ### Admin and moderation

  - Added admin dashboard, emergency overview, history, analytics, and deployment monitoring
  - Added admin users tab with listing and filters
  - Added ban/unban moderation
  - Added requester and volunteer side effects when users are banned

  ### Notifications

  - Added in-app notification infrastructure
  - Added FCM push notification support
  - Added notification devices, preferences, delivery tracking, cleanup jobs, and related tests

  ### News and announcements

  - Added admin announcement management
  - Added public news list and announcement detail pages
  - Added seeded initial announcements and deployment-safe migration behavior

  ### Android and CI

  - Added Android CI and e2e infrastructure
  - Added Android reset-password support
  - Added Android gathering areas, help request map, safety status, operational location, and sync
  improvements
  - Improved Android offline sync and pull-to-refresh behavior

  ## Notable fixes

  - Hardened guest help-request behavior
  - Made selected help-request fields optional where needed
  - Scoped login rate limiting by email and IP
  - Fixed profile name, DOB, location, and coordinate persistence issues
  - Fixed gathering-area provider failure, stale state, empty state, and map UX issues
  - Fixed location-sharing permission and consent edge cases
  - Fixed availability expiry and stale volunteer location handling
  - Fixed unsafe emergency fallback/demo request data
  - Fixed notification cleanup and stale delivery behavior
  - Stabilized backend integration, web e2e, and Android test flows

[...]

Issue #465, opened 2026-05-06
Release v0.2.0 from development to main


AlperKartkaya:
## Summary

  Prepare and release `v0.2.0` by merging the current `development` branch into `main`.

  This release includes the major feature and stability work completed after `v0.1.0`.

  ## Major features

  - Advanced help-request matching with multi-responder assignment support
  - Requester help-request current/history views with urgency, priority, and lifecycle metadata
  - Offline Android help request draft creation and sync recovery
  - Public active help-request visibility API and crisis/help-request maps
  - Web and Android gathering areas with map-based nearby results
  - Resilient gathering-area fallback states for provider/backend failures
  - Reusable map components and in-app map picker flows
  - Operational location snapshots and freshness checks for emergency workflows
  - User safety status tracking with backend persistence
  - Trusted safety circles and safety circle owner/member management
  - Offline safety check-in support
  - Admin dashboard with emergency overview, history, analytics, and deployment monitoring
  - Admin user listing, filtering, ban/unban moderation, and requester/volunteer ban side effects
  - In-app and FCM push notification infrastructure
  - Notification devices, preferences, delivery records, cleanup jobs, and related tests
  - Admin announcement management
  - Public news and announcement detail pages
  - Seeded initial announcements and deployment-safe migration behavior
  - Android CI/e2e test infrastructure
  - Android reset password, gathering areas, help request map, safety status, operational location, and
  sync improvements

  ## Major bug fixes

  - Hardened guest help-request behavior
  - Made selected help-request fields optional where needed
  - Scoped login rate limiting by email and IP
  - Fixed profile name, date of birth, location, and coordinate persistence flows
  - Fixed web/Android location tree and country-specific location parsing
  - Fixed gathering-area provider failure handling, stale states, empty states, and map UX issues
  - Fixed location-sharing permission and consent edge cases
  - Fixed availability expiry and stale volunteer location handling
  - Fixed effective availability calculation
  - Fixed unsafe emergency fallback/demo request data
  - Fixed safety status cleanup on account switch
  - Fixed notification cleanup and stale delivery behavior
  - Stabilized backend integration, web e2e, and Android test flows

  ## Release checklist

  - [ ] Confirm `development` contains all intended `v0.2.0` changes
  - [ ] Confirm no release-blocking regressions remain
  - [ ] Backend unit/integration tests pass or blockers are documented
  - [ ] Web build/e2e checks pass or blockers are documented
  - [ ] Android compile/tests pass
[...]

PR #481, opened 2026-05-07
Fix/release conflicts and security alerts

kckagancan:
## Summary

Resolves release merge dependency conflicts between `development` and `main`.

## Changes

- Preserves development web dependencies such as Leaflet and Playwright.
- Preserves backend Firebase/notification dependencies.
- Keeps Next.js security update.
- Forces PostCSS resolution to 8.5.10 so transitive Next/PostCSS usage no longer resolves to the vulnerable version.
- Updates backend dependency tree so `express-rate-limit` resolves with patched `ip-address`.
- Keeps backend top-level `uuid@14`.

## Verification

- Web audit: 0 vulnerabilities
- Web build: passed
- Backend unit tests: passed
- Android debug build: passed
- Android e2e unit tests: passed

## Known non-blocking audit note

Backend still reports low-severity transitive `@tootallnate/once` through `firebase-admin`.
This is not force-fixed because `npm audit fix --force` attempts a breaking Firebase Admin downgrade.

Related to PR #466 
Related to issue #465


AlperKartkaya (changes requested):
Thanks for the fix. Before merging, it seems to me there are a few blocking issues:

  - HIGH — web/package-lock.json is platform-pruned / incomplete
      - The branch removed optional native packages needed by Tailwind/Lightning CSS/Sharp on non-Linux
        platforms.
      - Evidence:
          - npm ci with local npm 11 fails: missing many optional packages from lockfile.
          - CI-like npm 10 install succeeds, but npm run build fails on macOS:
              - Cannot find module '../lightningcss.darwin-arm64.node'
      - This means the branch can break local development/builds on macOS, and the lockfile is not
        healthy.
      - Fix: regenerate web/package-lock.json without omitting optional/platform dependencies, ideally
        with the repo’s CI Node/npm version, then rerun install/build.

  Non-blocking notes:

  - Backend unit tests pass: 343/343
  - Backend high-severity audit passes: 0 high vulnerabilities
  - Web audit reports 0 vulnerabilities
  - Backend still has 8 low-severity audit findings via firebase-admin transitive deps. Not necessarily
    merge-blocking unless the release requires zero audit findings.

Also,

  For a better release-safe version, consider running:

```
  cd web
  rm -rf node_modules .next
  npm install
  npm run build
  npm audit --audit-level=high
```

  Then **commit**:

  - web/package-lock.json
  - and only web/package.json if it intentionally changed.

AlperKartkaya (approved):
Thanks for the check, now it looks OK to merge from my side.

PR #491, opened 2026-05-07
Fix Android lifecycle timestamp timezone display

### LINKED ISSUE #405: [Bug] Timezone is GMT+0 only.
We better update the timezone based on the location of the user, or take it from the user when creating profile (it may be silly if the user has already specified their location), or simply make use of the time of the phone - if simpler.

delibas-rojhat:
**Closes**

- Closes #405

**Summary**

Fixes Android lifecycle timestamp display so backend UTC timestamps are shown in the device's local timezone instead of appearing as GMT+0.

**Changes**

- Parses lifecycle timestamps as instants before formatting.
- Displays request lifecycle times using the phone's local timezone.
- Covers guest online refresh flows where server timestamps replace local offline timestamps.
- Adds timezone-focused unit tests.
- Keeps existing mapping tests deterministic by pinning their timezone.


kckagancan (approved):
Looks good. Merging.

PR #501, opened 2026-05-07
fix: Android location permission bootstrap and privacy defaults

### LINKED ISSUE #486: [Bug] Location permission missing - privacy related
When the app is opened for the first time on mobile, 
- we better prompt user to give location permission, auto-updating privacy status details in the settings updating "saved location visibility" accordingly.
- it is also **NOT** preferable that we set "profile visibility" and "health information visibility" - at the page privacy & security under settings on mobile - private by default

berat-sayin:
Related to the issue [#486](https://github.com/bounswe/bounswe2026group6/issues/486).

## Summary
- prompt for foreground location permission on first Android launch
- sync the initial permission result into saved location visibility defaults
- default profile and health visibility to `EMERGENCY_ONLY` instead of `PRIVATE`
- add backend migration/runtime defaults and focused tests

### DIFF
--- a/android/app/src/main/java/com/neph/MainActivity.kt
+++ b/android/app/src/main/java/com/neph/MainActivity.kt
@@ -2,11 +2,13 @@ package com.neph
 
 import android.Manifest
 import android.app.Activity
+import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.BackHandler
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.material3.AlertDialog
@@ -33,6 +35,7 @@ import com.neph.features.availability.data.AvailabilityRepository
 import com.neph.features.auth.data.AuthSessionStore
 import com.neph.features.operationallocation.data.OperationalLocationRepository
 import com.neph.features.operationallocation.data.OperationalLocationUpdater
+import com.neph.features.profile.data.DeviceLocationProvider
 import com.neph.features.profile.data.ProfileRepository
 import com.neph.features.notifications.data.PushTokenSync
 import com.neph.features.requesthelp.data.RequestHelpRepository
@@ -40,9 +43,18 @@ import com.neph.features.safetystatus.data.SafetyStatusRepository
 import com.neph.navigation.AppNavGraph
 import com.neph.navigation.Routes
 import com.neph.ui.theme.NephTheme
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.launch
 
 class MainActivity : ComponentActivity() {
+    private val initialLocationPermissionLauncher = registerForActivityResult(
+        ActivityResultContracts.RequestMultiplePermissions()
+    ) { grants ->
+        val granted = DeviceLocationProvider.hasLocationPermission(this) || grants.values.any { it }
+        syncLocationPermissionPrivacy(granted)
+        requestNotificationPermissionIfNeeded()
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         NephAppContext.initialize(applicationContext)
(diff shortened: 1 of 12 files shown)


kckagancan (changes requested):
Thanks, this is moving in the right direction and the touched areas match #486: Android first-launch location permission, mobile privacy defaults, backend default visibility, and tests.

I still want to request changes before approval because the current implementation has two privacy/behavior issues.

### 1. Do not override an existing saved location visibility during permission bootstrap

`ProfileRepository.syncPrivacyDefaultsForLocationPermission()` currently fetches the remote profile, but then always patches:

    locationVisibility = defaultLocationVisibilityForPermission(locationPermissionGranted)

This can silently overwrite an existing user choice. For example:

- The user previously set saved location visibility to `PRIVATE`
- The app is opened on a new install/device
- Location permission is granted
- The bootstrap can change saved location visibility to `EMERGENCY_ONLY`

The reverse can also happen if permission is denied and the user previously had `PUBLIC` or `EMERGENCY_ONLY`.

Permission result should influence the default only when saved location visibility is unset/uninitialized. It should not overwrite an explicit existing backend value.

Please update the logic and add a focused test showing that an existing `locationVisibility` value is preserved during permission bootstrap.

### 2. Logged-out first launch does not actually auto-sync the permission-derived default after login

On first launch, `syncPrivacyDefaultsForLocationPermission()` stores the local permission hint, but if there is no token it returns without patching the backend.

I do not see a later login/verify-email hook that applies this stored hint once the user becomes authenticated. As a result, the common first-install flow can be:

1. User opens the app while logged out
2. User grants location permission
3. Token is blank, so only the local hint is stored
4. User logs in
5. Backend still returns `locationVisibility = PRIVATE`
6. Privacy & Security still shows `PRIVATE`

That does not fully satisfy #486’s expectation that the initial permission result updates the saved location visibility accordingly.

Please apply the stored permission hint after authentication, or otherwise ensure that the permission-derived default is synced once a token exists.

The backend profile/health default changes look generally fine, and CI is green, but these two Android privacy/default-flow issues should be fixed before approval.

berat-sayin:
## Review Fixes

- preserve existing `locationVisibility` during Android permission bootstrap
- add `locationVisibilityInitialized` to profile responses so Android can distinguish explicit backend values from defaults
- defer first-launch permission-derived privacy sync while logged out
- apply pending permission hint after login and email verification
- add focused Android/backend unit tests for the reviewed cases

kckagancan (changes requested):
Thanks, the first requested change looks fixed now.

The existing saved `locationVisibility` is no longer blindly overwritten during permission bootstrap. The new `locationVisibilityInitialized` flag lets Android distinguish explicit backend values from defaults, and the focused tests cover preserving an existing initialized value.

However, I think the second requested change is still not fully fixed for the normal login flow.

`syncPrivacyDefaultsForLocationPermission()` stores the pending permission hint in `ProfileRepository` preferences when the user is logged out. But `AuthRepository.login()` later calls `ProfileRepository.clearProfile()` before calling `ProfileRepository.syncPendingLocationPermissionPrivacyHintIfNeeded()`.

Since the pending hint is stored in the same profile preferences, `clearProfile()` can remove:

- `locationPermissionPrivacyHint`
- `pendingLocationPermissionPrivacyHintSync`

before the pending sync has a chance to run.

So the common flow can still fail:

1. User opens the app while logged out
2. User grants location permission
3. Pending permission hint is stored locally
4. User logs in normally
5. `ProfileRepository.clearProfile()` clears the pending hint
6. `syncPendingLocationPermissionPrivacyHintIfNeeded()` sees no pending hint and does nothing

Please preserve the pending permission hint across `clearProfile()` during login, move it to a separate prefs store, or read/re-apply it before clearing the profile cache.

Please also add a focused test for the logged-out permission hint -> normal login -> pending privacy sync path.

berat-sayin:
## Review Fix

- move the first-launch location permission hint and pending privacy sync marker into a separate prefs store
- keep `ProfileRepository.clearProfile()` limited to profile cache state so normal login no longer erases the pending permission-derived privacy sync
- add a focused Android unit test for logged-out permission hint -> normal login profile clear -> pending sync survives

kckagancan (approved):

PR #525, opened 2026-05-07
feat: add demo seed migration

### LINKED ISSUE #383: [Feature] Add demo-ready seed migrations with meaningful NEPH data

gulcetahtasiz:
## Summary

Adds demo-ready NEPH seed data as a backend SQL migration instead of a standalone seed script.

This migration creates realistic demo users, profiles, volunteers, help requests, assignments, safety status data, notifications, announcements, and related location records for demo/testing scenarios.

## Changes

- Added SQL migration for demo-ready NEPH data
- Removed the old standalone `seed-demo-data.js` script
- Added 8 meaningful demo accounts: admin, requesters, and volunteers
- Added realistic Istanbul-based residential, operational, and request locations
- Added volunteer availability and expertise data
- Added active, assigned, resolved, and category-diverse help requests
- Added assignment, safety circle/status, notification, and announcement demo records
- Updated README demo seed documentation

## Notes

- Demo coordinates are realistic Istanbul land locations
- Seed data uses deterministic `demo_*` identifiers
- Migration uses conflict handling to stay safe for repeated/demo setup runs

## Testing

Ran:

`npm run migrate`

`npm run test:unit`

Result: backend unit tests passed.

Closes #383


kckagancan (changes requested):
Thanks for reworking this issue. The new implementation is much closer to what we wanted because the demo seed data is now represented as SQL migration data instead of the old standalone `seed-demo-data.js` script. The dataset itself is also meaningful: it includes demo admin/requester/volunteer accounts, profiles, locations, volunteer availability, help requests, assignments, safety status data, notifications, and announcements.

However, I do not think this PR is merge-ready yet. There are a few important scope and behavioral issues that should be fixed before approval.

### 1. Demo data is now inserted unconditionally by normal migrations

The biggest problem is that the new demo seed file lives directly under:

`backend/migrations/20260507_153000__seed_demo_ready_neph_data.sql`

Our migration runner applies every `.sql` file under `backend/migrations` during the normal migration flow. That means every environment that runs regular backend migrations will automatically receive demo users, demo help requests, demo announcements, demo notifications, etc.

This conflicts with the issue requirement that regular migrations should still run normally without forcing demo data into every environment.

The issue was reopened because we wanted mock/demo data as SQL migration files instead of a JS seed script, but that does not necessarily mean demo data should be part of the normal production migration path.

Please change this so demo seed data is applied only through an explicit demo-seed flow, for example:

- move demo seed SQL to a separate folder such as `backend/demo-migrations/` or `backend/migrations/demo/`
- add a dedicated command such as `npm run migrate:demo` or `npm run seed:demo`
- require an explicit guard such as `ENABLE_DEMO_SEED=true`
- make sure normal `npm run migrate` / `start:with-migrations` does not automatically insert demo data into every database

The important point is: demo data should be easy to apply for demo/evaluation, but not forced into every environment by default.

### 2. Current conflict handling is idempotent, but not fully safe

The migration uses deterministic `demo_*` IDs and `ON CONFLICT ... DO UPDATE`, which prevents duplicate records when the seed is run multiple times. That part is good.

However, the current approach can overwrite existing rows if a row with the same deterministic demo ID already exists. For example, if `demo_user_*`, `demo_profile_*`, `demo_request_*`, etc. already exist for any reason, the migration updates them directly.

This is probably fine for a clean demo database, but the issue also says existing production data must not be corrupted. A safer approach would be better, especially if demo seed remains reusable.

Please consider safer behavior such as:

- only updating records that are clearly demo-owned
- failing loudly if a deterministic demo ID exists but does not match the expected demo email/data
- using a dedicated `demo_*` namespace consistently and avoiding broad overwrites
- avoiding changing existing non-demo records through conflict updates

At minimum, the seed flow should be explicitly guarded so this risk is not present in normal migration runs.

### 3. Demo data creates multiple active/open help requests for the same requester

`demo_user_requester_2` owns both:

- `demo_request_assigned_food_water`
- `demo_request_active_search_rescue`

Technically this may not violate the current backend schema, but product-wise we have discussed the idea that a user should not normally have multiple active help requests at once. For demo purposes, this may look confusing in “My Help Requests” or during the user journey.

Please either:

- move one of these requests to another requester, or
- confirm that this is intentional and matches the current product behavior

A cleaner demo dataset would probably avoid one requester having multiple active/open requests unless that scenario is specifically being demonstrated.

### 4. Demo coverage is good, but could be slightly more complete for admin/history views

The current dataset has `PENDING`, `ASSIGNED`, and `RESOLVED` help requests, which is already useful.

However, admin/demo pages would be even more representative if we also had at least one of:

- `IN_PROGRESS` request
- `CANCELLED` request
- older historical request

This is not a hard blocker, but since the issue explicitly mentions overview/history/insights pages being meaningful, adding one cancelled or in-progress case would improve the demo value.

### 5. Please make sure the final design matches the reopened issue decision

The reopened issue comment says we need migration files of mock data instead of the script. This PR correctly moves away from the JS script, but we still need to preserve operational safety.

Expected final behavior should be:

- demo data is stored as SQL seed/migration-style data
- demo data can be applied repeatably/idempotently
- demo data is easy to apply on EC2 for the final demo
- normal migrations do not accidentally seed demo data into every environment
- existing real data is not overwritten or corrupted
- README clearly explains the exact command/flow

gulcetahtasiz:
Thanks for the review, updated the PR

Changes made:
- Moved demo seed SQL out of normal `backend/migrations/` into `backend/demo-migrations/`
- Added explicit guarded demo seed command: `ENABLE_DEMO_SEED=true npm run seed:demo`
- Kept normal `npm run migrate` / `start:with-migrations` from inserting demo data automatically
- Added safety checks before applying demo-owned seed records
- Expanded the demo dataset to 24 people:
  - 1 admin
  - 9 requesters
  - 9 volunteers
  - 5 additional community residents
- Added richer demo coverage with 11 help requests across `PENDING`, `ASSIGNED`, `IN_PROGRESS`, `RESOLVED`, and `CANCELLED`
- Added realistic Istanbul-based locations, health/profile data, volunteer availability, assignments, safety statuses, notifications, and announcements
- Updated README with the explicit demo seed flow

Verified locally:
- normal migrations do not apply demo seed data
- guarded demo seed runs only with `ENABLE_DEMO_SEED=true`
- backend unit tests pass

kckagancan (approved):
Great work. Merging.

PR #529, opened 2026-05-08
feat(android): date of birth calendar picker with auto-formatted manual entry

erinc00:
Replaces the plain text DOB field on Complete Profile and Edit Profile with a new `DateInput` component.

- Tapping the calendar icon opens a native `DatePickerDialog` with future dates disabled and the existing value preselected.
- Manual typing accepts digits only and inserts dashes automatically via `VisualTransformation`, so cursor position never shifts (fixes the digit-shift bug where typing `200405` would produce `2004-50`).
- Both input methods emit the same `YYYY-MM-DD` string, keeping the existing save/validate logic untouched.

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -46,6 +46,7 @@ import com.neph.ui.components.display.SaveActionBar
 import com.neph.ui.components.inputs.AppDropdown
 import com.neph.ui.components.inputs.AppTextArea
 import com.neph.ui.components.inputs.AppTextField
+import com.neph.ui.components.inputs.DateInput
 import com.neph.ui.components.selection.AppCheckbox
 import com.neph.ui.layout.AuthScaffold
 import com.neph.ui.map.MapPickerDialog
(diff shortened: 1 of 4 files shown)


kckagancan (changes requested):
Thanks for the implementation. The new `DateInput` component is a good direction and the picker path mostly preserves the existing `YYYY-MM-DD` save/validation flow. Android CI also appears to pass.

However, I do not think this fully satisfies the linked DOB picker issue yet.

The issue scope says Date of Birth should be selected from a calendar picker on both Android profile flows, and the acceptance criteria says:

- “Tapping Date of Birth opens a native calendar picker.”
- “Manual date typing is no longer required.”
- Behavior should be consistent in both Complete Profile and Edit Profile.

In the current implementation, tapping the calendar icon opens the native `DatePickerDialog`, but tapping the Date of Birth field itself still behaves like a normal editable text field and brings the user into manual typing. That means the primary field interaction still does not match the acceptance criteria as written; the PR narrows the behavior to “tap the calendar icon,” while the issue asks for the DOB field to open the picker.

Please update `DateInput` so that the Date of Birth field itself opens the picker in both Complete Profile and Edit Profile. If the intended product decision is to keep hybrid manual typing + calendar icon, then the issue/PR description should be updated accordingly, but as currently written this should not close the issue.

Suggested fix direction:

- Make the DOB input read-only and open the picker when the field is tapped, or otherwise attach the picker-opening behavior to the whole DOB input area, not only the trailing icon.
- Keep `datePicker.maxDate` so future dates remain disabled.
- Keep existing valid DOB preselection behavior.
- Keep emitting backend-compatible `YYYY-MM-DD`.

Non-blocking note: `AppTextField` now has an `onClick` parameter, but this PR does not actually use it for the new `DateInput`, so that change currently looks unused/out-of-scope unless it is part of the intended fix.

erinc00:
The onClick parameter has been removed from AppTextField — it was added during early exploration of this approach but was never actually used once DateInput became its own standalone composable.

Regarding the "field itself should open the picker" feedback: the current design is intentional. DateInput supports both manual YYYY-MM-DD entry (with auto-inserted dashes via VisualTransformation) and the calendar icon to open the native picker. This hybrid approach was the product decision behind this PR — it mirrors the web DateInput component and gives users the flexibility to type directly or use the picker. The issue description and PR will be updated to clarify this scope.

kckagancan (approved):

Issue #527, opened 2026-05-08
[BUG] Use calendar picker for Date of Birth on Android profile screens


erinc00:
Description:
Date of Birth is currently entered as free text on mobile, which is error-prone. On Android profile flows, this field should be selected from a calendar picker.

Scope:

In Complete Profile, Date of Birth should open a native date picker.
In Edit Profile, Date of Birth should open a native date picker.
Save value in existing backend-compatible format (YYYY-MM-DD).
Prevent selecting future dates.
If a DOB already exists, preselect it when opening the picker.

Acceptance Criteria:

Tapping Date of Birth opens a native calendar picker.
Future dates cannot be selected.
Saved value format remains compatible with current API.
Behavior is consistent in both Complete Profile and Edit Profile.

Out of Scope:
Web Date of Birth UI changes.

Issue #528, opened 2026-05-08
[BUG] Remove Profession from profile on Android and Web, keep only first-aid knowledge option


erinc00:
Description:
Remove the Profession field from profile flows on both mobile and web. Keep only one first-aid knowledge option. Rename existing First Aid label to: Do you know first aid?

Scope:
Remove Profession input/display from Android and Web profile UIs.
Keep only a single first-aid checkbox/toggle.
Rename label to Do you know first aid?
Remove client-side profession patch/update usage where applicable.
Keep backend compatibility (profession can remain in backend model if needed, but not used in UI).

Acceptance Criteria:
Profession is not visible on Android and Web profile screens.
Profession is removed from Complete Profile and Edit Profile flows.
User can only set first-aid knowledge option.
Label is exactly Do you know first aid?
Profile save/update flow works without regressions.
Existing profession data does not break profile rendering or editing.

Test Checklist:
New user can complete profile without Profession field.
Existing user can edit profile without Profession field.
First-aid selection persists after save.
Behavior matches between Android and Web.

PR #530, opened 2026-05-08
Fix/profile remove profession keep first aid, add web DOB picker

erinc00:
### Summary

This PR cleans up the profile forms on both Android and Web by removing the unused Profession field and improving the First Aid and Date of Birth UX.

### Changes

**Remove Profession**
- Removed Profession dropdown from Complete Profile and Edit Profile on Android and Web
- Removed `patchMyProfession` API calls from both web forms
- `professionOptions` still exists in `profileOptions.ts` but is no longer used in the UI

**First Aid**
- Renamed expertise label from `"First Aid"` to `"Do you know first aid?"` for clarity
- Legacy values stored as `"First Aid"` are normalized to the new label on load (Android + Web)

**Date of Birth — Web** *(aligned with Android implementation in #527)*
- Replaced `<input type="date">` with a new `DateInput` component
- Supports manual `YYYY-MM-DD` entry with auto-inserted dashes
- Calendar icon opens the native date picker
- Future dates are rejected on manual entry; `max` constraint applied to the native picker

**Cleanup**
- Removed two placeholder helper text messages from Android profile screens (`"Document upload is still unavailable..."` and `"This is your home/neighborhood location..."`)

---

### Testing

- Android: `compileDebugKotlin` BUILD SUCCESSFUL
- Web: `tsc --noEmit` clean

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -392,7 +392,6 @@ fun CompleteProfileScreen(
             }
 
             Text("Residential Location", style = MaterialTheme.typography.titleMedium)
-            HelperText(text = "This is your home/neighborhood location. It is not automatically updated by GPS.")
 
             if (locationLoading) {
                 HelperText(text = "Loading location options...")
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -33,7 +33,6 @@ import com.neph.features.profile.data.locationData
 import com.neph.features.profile.data.normalizeDateOfBirth
 import com.neph.features.profile.data.normalizePhoneParts
 import com.neph.features.profile.data.parseListField
-import com.neph.features.profile.data.professionOptionsFor
 import com.neph.features.profile.data.sanitizeDecimalInput
 import com.neph.features.profile.data.splitFullName
 import com.neph.features.profile.data.toEditableString
@@ -96,7 +95,6 @@ fun CompleteProfileScreen(
     var district by rememberSaveable { mutableStateOf(existingProfile.district.orEmpty()) }
     var neighborhood by rememberSaveable { mutableStateOf(existingProfile.neighborhood.orEmpty()) }
     var extraAddress by rememberSaveable { mutableStateOf(existingProfile.extraAddress.orEmpty()) }
-    var profession by rememberSaveable { mutableStateOf(existingProfile.profession) }
     var expertise by rememberSaveable { mutableStateOf(existingProfile.expertise) }
     var loading by rememberSaveable { mutableStateOf(false) }
     var error by rememberSaveable { mutableStateOf("") }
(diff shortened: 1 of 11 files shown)


kckagancan (changes requested):
Thanks for the cleanup work. The PR is moving in the right direction: Profession is removed from the visible Android/Web profile UIs, the First Aid label is updated, legacy `"First Aid"` values are normalized, and CI appears to be green.

However, I do not think this fully satisfies the linked profile cleanup issue yet, and there is also a web DOB regression risk introduced by the extra DateInput change.

### 1. Android still performs profession updates

Issue #528 explicitly includes:

> Remove client-side profession patch/update usage where applicable.

The web forms removed `patchMyProfession`, but Android still keeps profession in the sync flow. `ProfileRepository.syncProfile()` still sends a PATCH request to `/profiles/me/profession`, and Complete Profile still carries the hidden `profession` state into the profile object even though the UI no longer lets the user edit it.

This means the field is removed visually, but Android still depends on and updates the profession endpoint during profile save. That does not match the issue scope.

Please remove the Android-side profession patch/update usage as well, or clearly justify why Android must keep calling it despite the issue saying client-side profession update usage should be removed.

Expected direction:

- Remove the `/profiles/me/profession` PATCH from Android profile save if Profession is no longer used by the client.
- Stop carrying hidden `profession` form state in Complete Profile unless there is a deliberate compatibility reason.
- Keep backend model compatibility if needed, but do not keep UI save flows dependent on the profession endpoint.

### 2. Web DateInput can save partial/non-YYYY-MM-DD DOB values

This PR also replaces web `<input type="date">` with a custom text-based `DateInput`. The component emits partial formatted values while typing, such as:

- `1998`
- `1998-01`
- `1998-01-`

That is fine for typing UX, but the save flows do not strictly validate `YYYY-MM-DD` before sending the value to the backend.

In Complete Profile, the validation uses `new Date(form.dateOfBirth)`, which is not strict enough for enforcing the existing API-compatible `YYYY-MM-DD` format. In Edit Profile, the save flow patches `dateOfBirth: profile.dateOfBirth || null` without a strict DOB format check.

This can regress profile save/update behavior because a user can leave a partial date in the custom text input and attempt to save a backend-incompatible value.

Please add strict date validation before save on both web profile flows, for example:

- require `^\d{4}-\d{2}-\d{2}$`
- reject invalid calendar dates
- reject future dates
- only send a full valid `YYYY-MM-DD` value or `null` where the existing API contract allows it

Alternatively, make `DateInput` only commit full valid dates to parent state, but the current partial-emitting behavior needs matching save validation.

Once the Android profession update path is removed and the web DOB save validation is made strict, this should be much safer to approve.

kckagancan (changes requested):
Thanks for the follow-up fix. The main Android profession blocker is now addressed: the hidden Complete Profile profession state was removed, and Android no longer patches `/profiles/me/profession` during `ProfileRepository.syncProfile()`.

However, I still do not think this is ready to approve because the new web DOB validation is not strict enough for the custom text-based `DateInput`.

The current validation checks:

- `^\d{4}-\d{2}-\d{2}$`
- `new Date(value)` is not `NaN`
- date is not in the future

This still allows invalid calendar dates such as `2023-02-31`, because JavaScript `Date` parsing can normalize invalid dates instead of rejecting them. Since this PR replaces the browser-native `<input type="date">` with a text input, users can now type values that the old native date input would not allow.

Please update the web DOB validation in both Complete Profile and Edit Profile so it verifies that the typed date is a real calendar date, not just a parseable string. For example:

- match `YYYY-MM-DD`
- split into year/month/day
- construct a date
- verify the constructed UTC/local year, month, and day exactly match the original input
- reject future dates

This should prevent cases like:

- `2023-02-31`
- `2024-13-01`
- `2024-00-10`
- `2024-04-00`

kckagancan (approved):

PR #553, opened 2026-05-09
fix: stabilize Android resource map UX

### LINKED ISSUE #542: [Bug] Missing "show my location" button on the maps
For both web and mobile, both the help request map and gathering areas map lack the following button: 

Image

It is not an optional feature to add, instead it is a real requirement making it sort of impossible for the user to find their location on the map, when this feature is missing.

### LINKED ISSUE #543: [Bug] Cannot navigate on the maps on mobile
- When tried to navigate to a specific are on the map, it is hardly possible to do so, as the page scrolling is not disabled. 
#### Expected behaviour
- When the user touches the map regardless of their purpose, page scrolling should be disabled.

### LINKED ISSUE #548: [Bug][Android] Help Request Map Details Collapse Horizontally
#Summary

Help Request Map detail cards break on smaller screen sizes and collapse text content horizontally.

Priority/type labels are rendered character-by-character vertically instead of wrapping/layouting correctly, making the card difficult to read and visually broken.

Buttons remain visible, but the left content section appears to shrink/collapse incorrectly.

---

#Expected Behavior

Card content should remain readable and responsive across supported screen sizes.

Priority/type text should wrap normally without collapsing vertically.

---

#Actual Behavior

Text content collapses into a narrow vertical column on smaller/mobile layouts, breaking the card UI and readability.

Image

kckagancan:
### Summary

This PR fixes Android resource map UX regressions observed after viewport-based marker discovery was added.

The changes focus on making the Android Gathering Areas and Help Request Map screens demo-safe and usable on real devices by improving map gestures, map state messages, current-location behavior, loading behavior, and responsive request card layout.

#### Shared Android map behavior

- Fixed mobile map gesture handling so dragging or pinching inside the map does not scroll the parent page.
- Kept Leaflet/WebView touch events working while preventing parent scroll interception.
- Preserved existing WebView lifecycle, tile loading, initialization handling, and JavaScript marker updates.

#### Gathering Areas map

- Removed duplicated helper/status messages such as repeated `Zoom in to see resources in this area.`
- Improved zoomed-out, loading, empty, provider-error, and retry state separation.
- Moved current-location actions into map overlay controls and improved its recenter zoom behavior.
- Avoided misleading “No Gathering Areas Found” messaging when the actual state is zoomed out or provider unavailable.
- Improved provider fallback/stale metadata handling so provider failures are not treated as true empty results.
- Reduced noisy loading behavior during viewport panning.

#### Help Request Map

- Added or restored a current-location shortcut for the Android Help Request Map where it was missing.
- Removed misleading filter-empty text when the map is actually zoomed out.
- Reduced duplicated map helper/loading messages.
- Improved viewport update behavior so panning does not feel like a full screen reload.
- Fixed small-screen request detail/list card layout so text does not collapse horizontally.
- Preserved request filters, marker selection, `Get Directions`, and `Open in Map`.

### Related issues

Closes #543  
Closes #548  
Closes #542  

### Testing

- Verified Android resource map state messages are no longer duplicated.
- Verified zoomed-out state is not shown as filter-empty or true-empty.
- Verified map touch gestures no longer cause parent page scrolling.
- Verified Gathering Areas `Use Current Location` recenters with a closer local zoom.
- Verified Help Request Map current-location shortcut behavior.
- Verified Help Request Map request cards remain readable on small screens.
- Verified marker selection, filters, `Get Directions`, and `Open in Map` still work.
- Verified map picker flows were not intentionally changed.

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
+++ b/android/app/src/main/java/com/neph/features/auth/presentation/CompleteProfileScreen.kt
@@ -530,11 +530,20 @@ fun CompleteProfileScreen(
                     initialLongitude = mapPickerSelection?.longitude,
                     centerLatitude = mapCenterLatitude,
                     centerLongitude = mapCenterLongitude,
+                    showCenterOnCurrentLocation = true,
                     centerActionLoading = mapCenterLoading,
                     centerActionMessage = mapCenterInfo,
                     loading = mapPickerLoading,
                     onDismiss = { if (!mapPickerLoading) mapPickerOpen = false },
-                    onConfirm = ::handleMapSelection
+                    onConfirm = ::handleMapSelection,
+                    onCenterOnCurrentLocation = {
+                        if (mapLocationPermissionRequester.refreshPermissionState()) {
+                            captureCurrentLocationForProfileMap()
+                        } else {
+                            mapCenterLoading = true
+                            mapLocationPermissionRequester.requestPermission()
+                        }
+                    }
                 )
             }
 
--- a/android/app/src/main/java/com/neph/features/gatheringareas/data/GatheringAreasRepository.kt
+++ b/android/app/src/main/java/com/neph/features/gatheringareas/data/GatheringAreasRepository.kt
@@ -1,5 +1,7 @@
 package com.neph.features.gatheringareas.data
 
+import android.util.Log
+import com.neph.BuildConfig
 import com.neph.core.network.ApiException
 import com.neph.core.network.JsonHttpClient
 import kotlinx.coroutines.withTimeoutOrNull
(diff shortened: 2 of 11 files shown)


erinc00 (approved):
Overall this looks good to me, and I’m comfortable approving it.

The Android map UX improvements here are meaningful: gesture handling is cleaner, viewport updates look less disruptive, state messaging is better separated, and the current-location flow is more usable on both Gathering Areas and Help Request Map. I also like that the stale/provider-empty distinction is now handled more explicitly instead of being surfaced as a misleading true-empty state.

One note to call out: the PR description says map picker flows are unchanged / out of scope, but this diff also includes changes in MapPickerDialog and CompleteProfileScreen related to current-location centering. I do not see this as a blocker from the visible diff, but I think it is worth noting so the final PR description/scope stays accurate.

Approved.

PR #560, opened 2026-05-10
fix(android): allow non-com emails in auth validation

### LINKED ISSUE #558: [Bug] Mobile signup/login email field only accepts .com email addresses
Description
There is a bug on mobile during signup and login where the email input appears to accept only email addresses with a .com domain.

Users with valid email addresses using other domain extensions such as .net, .org, .edu, .co.uk, .io, .gov, etc. cannot proceed even though those addresses are valid.

This causes valid users to be blocked from registering or logging in on mobile.

Expected Behavior
The email field should accept any valid email format, not only .com addresses.

Actual Behavior
The field rejects or does not properly allow valid email addresses unless they end with .com.

Scope

* Mobile signup screen
* Mobile login screen

Impact
High user-facing issue. It blocks users with valid non-.com email addresses from accessing the app.

Notes
Likely caused by overly restrictive email validation logic on mobile.

erinc00:
Description  
This PR fixes mobile auth email validation that incorrectly blocked valid non-.com emails.

- Updates UI-side email validation to remove domain-extension bias.
- Ensures valid domains like .net, .org, .edu, .io, and .co.uk are accepted.
- Adds unit tests for valid non-.com emails and invalid formats.

Scope  
- Mobile auth validation utility only.
- No backend contract changes.

Closes #558

### DIFF
--- a/android/app/src/main/java/com/neph/features/auth/util/AuthValidation.kt
+++ b/android/app/src/main/java/com/neph/features/auth/util/AuthValidation.kt
@@ -1,7 +1,13 @@
 package com.neph.features.auth.util
 
-import android.util.Patterns
-
 fun isValidEmail(email: String): Boolean {
-    return Patterns.EMAIL_ADDRESS.matcher(email.trim()).matches()
+    val normalized = email.trim()
+    if (normalized.isEmpty()) {
+        return false
+    }
+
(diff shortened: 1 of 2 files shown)


kckagancan (approved):
Looks fine. Merging.

PR #563, opened 2026-05-10
fix: mobile account soft deletion and fresh re-signup handling

### LINKED ISSUE #554: [Bug] Account safe-deletion fails on mobile
#### Current behaviour
- When the user clicks "delete account" on mobile settings, operation fails with an error "Something went wrong".

#### Expected behaviour
- Account deletion should direct the user to login page, marking the user's account as deleted on the database, without actually deleting it.

**Important Note:**
- When the very same user attempts to create an account on the database side, everything should be handled carefully in a way that the previous record should not be lazily used again (like unmarking `is_deleted` field only).

Related to #504

berat-sayin:
Related to the issue [#554](https://github.com/bounswe/bounswe2026group6/issues/554).

This PR fixes account deletion behavior for mobile and tightens backend soft-delete identity handling.

### Changes

- Navigate mobile users to the login screen after successful account deletion.
- Clear google_id during backend soft deletion so deleted Google accounts do not keep owning the OAuth identity.
- Allow signup to create a fresh user row when the matching existing account is already soft-deleted.
- Add backend coverage for re-signup after deleted email/Google accounts.
- Update Android e2e expectation for the post-delete destination.


kckagancan (approved):
Great work. Looks fine. Merging.

PR #565, opened 2026-05-10
fix: enable mobile privacy location sharing save flow

### LINKED ISSUE #555: [Bug] The button "Save Privacy Settings" does not work
I noticed that at the page "privacy & security" under settings on mobile, "profile visibility" and "health information visibility" seems to remain private after even after creating a couple of help requests. I though it is a problem that we do not update these fields, but, after a more careful look, I noticed the following:
- Even I turn on the toggle "Share Current Location" at the page "privacy & security" under settings on mobile, the app did not allow me to save it! The button "Save Privacy Settings" did not work, giving the message attached, and it is not recoverable on mobile even you go to the profile page as prompted.

Image

 

Related to: [#486](https://github.com/bounswe/bounswe2026group6/issues/486) [#504](https://github.com/bounswe/bounswe2026group6/pull/504)

berat-sayin:
Related to the issue [#555](https://github.com/bounswe/bounswe2026group6/issues/555).

## Summary
- Fixed the mobile Privacy & Security save flow when enabling “Share Current Location”.
- The screen now requests location permission, captures the current device location, persists it to the profile location endpoint, then saves privacy settings.
- If location permission is denied or current location is unavailable, other privacy settings still save and location sharing remains disabled with a clear message.

### DIFF
--- a/android/app/src/main/java/com/neph/features/privacysecurity/presentation/PrivacySecurityScreen.kt
+++ b/android/app/src/main/java/com/neph/features/privacysecurity/presentation/PrivacySecurityScreen.kt
@@ -13,8 +13,12 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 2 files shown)


kckagancan (approved):
Looks fine. Merging.

Issue #557, opened 2026-05-09
[Bug] Improve Google social auth reliability on Android and Web


erinc00:
## Summary
Google social authentication had reliability and UX issues across Android and Web. This issue tracks the fixes needed to make login/signup behavior consistent, improve error visibility, and reduce configuration-related failures.

## Problem
- Android Google login/signup could appear unresponsive in some cases.
- Credential parsing and error handling were not explicit enough for debugging and user feedback.
- Web social auth button behavior and signup consent placement had UX inconsistencies.
- Environment/config mismatches could lead to Google console setup errors.

## Scope
- Harden Android Google login and signup flows.
- Improve credential validation and fallback behavior.
- Improve user-facing error messages for Google auth failures.
- Align Web signup/social auth UX.
- Keep unrelated config files out of this change set.

## Acceptance Criteria
- [ ] Android Google login flow uses robust credential handling and fallback behavior.
- [ ] Android Google signup flow validates credential type and token integrity.
- [ ] Android shows clear, actionable error messages for Google auth failures.
- [ ] Web social auth button rendering is stable and full-width.
- [ ] Web signup consent placement is consistent with intended UX.
- [ ] All changes compile successfully for Android debug build.

PR #566, opened 2026-05-10
fix(android): pass Google server client ID to release build

kckagancan:
Fixed android release workflow and added a secret to enable google sign up flow. 

Related to #557


erinc00 (approved):
Approved and merging

PR #570, opened 2026-05-10
feat: update request help health sharing consent

### LINKED ISSUE #568: [Feature] simplify request help form and add health info sharing consent
### Description
The Request Help form should be simplified and updated to better match the emergency flow and privacy expectations. The help type options will be reduced, risk and vulnerable group fields will become optional, and blood type will be removed from the form in favor of explicit consent-based profile health information sharing.

#### 1. Limit Required Help options
Update the `Required Help` field so that only the following options are available:

- First Aid
- Food & Water
- Shelter
- Search & Rescue

Remove the following options from the Request Help form:

- Fire Brigade
- Evacuation / Transport
- Security Support
- Other

#### 2. Make Risk Flags optional
The `Risk Flags` field should be optional.

- Users should be able to submit the form without selecting any risk flags.
- The UI should clearly indicate that this field is optional.
- Empty selections should be accepted by the backend.

#### 3. Make Vulnerable Groups optional
The `Vulnerable Groups` field should be optional.

- Users should be able to submit the form without selecting any vulnerable groups.
- The UI should clearly indicate that this field is optional.
- Empty selections should be accepted by the backend.

#### 4. Replace Blood Type field with health info sharing consent
Remove the `Blood Type` field from the Request Help form.

Add a checkbox for explicit consent:

> I agree to share my profile health information with the volunteer assigned to this request.

Behavior:

- The checkbox should be unchecked by default.
- If unchecked, profile health information must not be shared with volunteers.
- If checked, the user’s profile health information may be shared with the volunteer assigned to the request.
- Profile health information includes existing profile health fields such as blood type, medical history/conditions, chronic diseases, and allergies.

### Expected Behavior

- Users can submit a help request by selecting at least one allowed Required Help option.
- Risk Flags and Vulnerable Groups do not block submission when empty.
- Blood Type is no longer entered or edited from the Request Help form.
- Health information is shared only when the user explicitly opts in for that help request.
- Volunteer-facing views should only display profile health information when the user granted consent for that request.

### Implementation Notes

- Android Request Help form must be updated.
- Backend request validation and payload handling should be updated accordingly.
- Offline/draft handling on Android should continue to work with the new fields.
- Web currently does not appear to have a Request Help creation form, but any volunteer/admin/request detail views should respect the new health info sharing consent.
- Existing profile health information remains in the profile flow.

### Acceptance Criteria

- [ ] Required Help only shows First Aid, Food & Water, Shelter, and Search & Rescue.
- [ ] Removed help types cannot be selected in the Request Help form.
- [ ] Risk Flags is visibly marked as optional.
- [ ] Vulnerable Groups is visibly marked as optional.
- [ ] The form can be submitted with empty Risk Flags and empty Vulnerable Groups.
- [ ] Blood Type is removed from the Request Help form.
- [ ] A health information sharing consent checkbox is added and unchecked by default.
- [ ] Profile health information is not shared when the checkbox is unchecked.
- [ ] Profile health information is shared only with the assigned volunteer when the checkbox is checked.
- [ ] Backend validation accepts the updated payload.
- [ ] Android offline/draft request flow works with the updated form.

gulcetahtasiz:
## Summary
- Simplified Request Help help type options to First Aid, Food & Water, Shelter, and Search & Rescue.
- Made Risk Flags and Vulnerable Groups optional in the Request Help form.
- Removed Blood Type from the Request Help form and added explicit profile health info sharing consent.
- Added backend and Android offline support for request-level health info sharing consent.
- Updated assigned volunteer data flow so profile health details are only exposed when consent is granted.

## Changes
- Removed Fire Brigade, Evacuation / Transport, Security Support, and Other from Required Help options.
- Updated backend validation to accept only the remaining four help types.
- Updated Risk Flags and Vulnerable Groups labels to clearly show they are optional.
- Confirmed empty Risk Flags and Vulnerable Groups are accepted and sent as empty arrays.
- Added `share_profile_health_info_with_volunteer` to help requests.
- Removed Blood Type from Android Request Help form payload/UI.
- Added unchecked-by-default health info sharing checkbox.
- Updated Android Room schema/migration for the new consent field.
- Updated assigned request UI to show profile health details only when the requester opted in.
- Updated related unit/offline mapping tests.

## Tests
- `backend npm run test:unit -- --runTestsByPath tests/unit/modules/help-requests/validators.test.js tests/unit/modules/help-requests/service.test.js tests/unit/modules/help-requests/controller.test.js`
- `android ./gradlew :app:test`
- `git diff --check`

## Note
Backend integration tests were attempted, but local PostgreSQL setup was unavailable during Jest global setup.

Closes #568


kckagancan (changes requested):
Thanks for the implementation. The product direction is correct. However, this PR still leaves a few regressions caused by the enum/field transition. These should be fixed before approval.

### 1. Web Crisis Map does not recognize the new `search_rescue` canonical value

The backend now accepts `search_rescue`, and Android Request Help now sends `Search & Rescue` as `search_rescue`.

However, `web/src/app/crisis-map/page.tsx` still maps only old aliases such as `search_and_rescue` and `fire_brigade` to `SEARCH_AND_RESCUE`.

Because of that, any newly created `search_rescue` request will appear on the Web Crisis Map as `Other / Unknown`, and it will not match the `Search and Rescue` filter.

Please update the Web Crisis Map normalization so the new canonical value is handled:

- `search_rescue` → `SEARCH_AND_RESCUE`

Keep old aliases only as backward compatibility if needed:

- `search_and_rescue`
- `sar`
- `fire_brigade`
- `rescue`

### 2. Android Help Request Map has the same `search_rescue` normalization gap

`android/app/src/main/java/com/neph/features/helprequestmap/data/ActiveHelpRequestsRepository.kt` also does not recognize `search_rescue` in `normalizeRequestType`.

So newly created Search & Rescue requests can also show up as `Other / Unknown` on the Android Help Request Map and fail to match the Search and Rescue filter.

Please update Android map normalization as well:

- `search_rescue` → `CrisisRequestType.SEARCH_AND_RESCUE`

Again, keeping old aliases is fine for backward compatibility.

### 3. Test coverage should include the new canonical value

The current map tests still do not prove that `search_rescue` renders as Search and Rescue.

Please add/update tests so both Web and Android map parsing/filtering cover the new canonical value:

- Web active help request fixture with `type: "search_rescue"` should render as `Search and Rescue`.
- It should match the Search and Rescue filter.
- It should not appear as `Other / Unknown`.
- Android `ActiveHelpRequestsRepository.normalizeRequestType("search_rescue")` should return `SEARCH_AND_RESCUE`.

This is important because CI is green right now, but the green tests do not catch this regression.

### 4. Demo seed / fixtures should be checked for old help type values

Since the product decision is now to use `search_rescue` as the canonical value, please also check demo seed data and test fixtures for old values such as:

- `search_and_rescue`
- `fire_brigade`
- `evacuation_transport`
- `security_support`
- `other`

Old aliases can remain supported for backward compatibility, but newly seeded/demo-created data should preferably use the new canonical values so the demo environment reflects the current product behavior.

kckagancan (changes requested):
Thanks for the fixes. The previous enum-transition blockers are resolved now.
There is one remaining deployment/data issue before approval.

### Blocker: do not rely only on editing an already-versioned demo migration

This PR modifies the existing demo migration:

`backend/demo-migrations/20260507_153000__seed_demo_ready_neph_data.sql`

That is okay for a completely fresh database, but it will not reliably update any environment where this demo migration has already been applied.

The demo seed runner treats demo seed SQL files as migrations. Since applied migrations are tracked by `schema_migrations.file_name`, a previously applied file with the same path will be skipped on the next `ENABLE_DEMO_SEED=true npm run seed:demo`.

So if Neon/demo DB already ran the old demo seed, editing this existing SQL file will not update the old help type values there. The DB will keep whatever old seeded values were already inserted.

Please add a new follow-up demo migration instead, for example:

`backend/demo-migrations/20260511_XXXXXX__normalize_demo_help_request_types.sql`

That migration should update existing demo-owned help requests to the new canonical values, for example:

- convert old Search & Rescue aliases to `search_rescue`
- remove/replace removed demo help types such as `fire_brigade`, `evacuation_transport`, `security_support`, and `other`
- keep the update scoped only to demo data, e.g. `request_id LIKE 'demo_%'` and/or existing `[DEMO]` markers

It is fine to keep the edited seed file for brand-new fresh DBs, but existing seeded environments need a new migration file so the correction actually runs.

### Non-blocking polish

Assigned Request display formatting currently turns `search_rescue` into `Search Rescue`. That is not a blocker, but if we want product copy consistency, it could be displayed as `Search & Rescue` / `Search and Rescue`.

After adding the follow-up demo migration for already-seeded environments, I think this PR is safe to approve.

kckagancan (approved):

PR #579, opened 2026-05-10
fix(android): clear safe status when requesting help

### LINKED ISSUE #550: [Bug] When marked safe, it stays safe continually
#### Current behaviour
- When user marks themselves safe, it stays safe forever.

#### Expected behaviour 
- Safe mark should be gone when user clicks "I need help now"

kckagancan:
## Summary

This PR fixes issue #550 where a user remained marked as safe after tapping **I need help now** on Android.

## Changes

- Added Android-side logic to clear the existing safe status when an authenticated user starts the request-help flow.
- Updated the request-help action so the Home screen no longer keeps showing the user as safe after they request help.
- Preserved the existing offline-first behavior for safety status updates.
- Avoided changing backend API contracts or unrelated mobile flows.

## Expected Behavior

After this change:

1. User marks themselves safe.
2. Home screen shows the safe status.
3. User taps **I need help now**.
4. The previous safe status is cleared/replaced.
5. Home screen no longer displays the user as safe while they are requesting help.

## Testing

- Verified the Android request-help flow manually.
- Checked that the safe status is no longer persisted after starting a help request.
- Ensured existing safety status and request-help logic remain scoped to Android.

Closes #550

### DIFF
--- a/android/app/src/androidTest/java/com/neph/e2e/SafetyStatusRepositoryAndroidTest.kt
+++ b/android/app/src/androidTest/java/com/neph/e2e/SafetyStatusRepositoryAndroidTest.kt
@@ -95,6 +95,49 @@ class SafetyStatusRepositoryAndroidTest {
     }
 
     @Test
+    fun clearSafeStatusForRequestHelpReplacesSafeStatusAndSyncsNotSafe() = runBlocking {
+        SafetyStatusRepository.markSafe(
+            token = "access-token-1",
+            location = sampleLocation(),
+            shareLocationConsent = true
+        )
+
+        val state = SafetyStatusRepository.clearSafeStatusForRequestHelp("access-token-1")
+
+        assertEquals(SyncStatus.SYNCED, state.syncStatus)
+        assertEquals("not_safe", state.status)
+        assertFalse(state.shareLocationConsent)
+        assertFalse(state.hasLocation)
+        assertEquals("not_safe", fakeBackend.currentSafetyStatus().status)
+        assertNull(fakeBackend.currentSafetyStatus().location)
+        assertNull(latestSafetyStatusOperation())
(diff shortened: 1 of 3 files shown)


AlperKartkaya:
It maybe is an edge case, but I noticed that when using the app as a guest user, the app throws the user back to the login page when the user clicks "I'm safe".

kckagancan:
I will look into that thanks. @AlperKartkaya

AlperKartkaya:
> I will look into that thanks. @AlperKartkaya

You're welcome. Here's a bit more feedback:

1. When the user attempts to create a help request, but then goes back not filling it, or cancels their help request that they filled completely, the status field remains as "Help requested" as  in the image below.

Screenshot 2026-05-10 at 23 53 18


2. Also, "status unknown" looks like a bug or some sort of a connection lost. Can we change it to something like "All services operational." or "Hope you're fine!", or "You look awesome!", with a light-greenish background? 

Screenshot 2026-05-11 at 00 04 23


3. Last but not least, I think, marking as safe should trigger canceling the existing help request - if there is - with an explicit warning that says "You are marking yourself safe, this will cancel your existing help request. Are you sure?". We should warn the user like that only if they really have an existing help request, for sure.

kckagancan:
Good catch on the first point @AlperKartkaya . The current PR clears `safe` by setting the safety status to `not_safe`, which makes Home show `Help requested` even if the user backs out before creating a real help request. I adjusted this so the safe badge is cleared without falsely implying that a request exists. Can you review the PR and merge if everythings fine now?

For points 2 and 3, I think those should be separate follow-up issues:
- Renaming/restyling `Status unknown` is a broader UI polish decision.
- Marking safe canceling an existing help request requires a proper confirmed flow and request cancellation/status-sync behavior, so it is larger than this fix.

erinc00 (approved):
Looks good now. Approved and merging.

PR #582, opened 2026-05-11
feat(android): improve home safety status UX

### LINKED ISSUE #581: [FEATURE] Improve Home safety status UX and handle active help requests when marking safe
## Description

Improve the Android Home safety status UX.

Currently, Home can show **"Status unknown"**, which feels technical and may look like a bug. Also, if a user has an active help request and then taps **I'm safe**, the app should handle that conflict explicitly instead of leaving both states ambiguous.

## Expected Behavior

- Replace **"Status unknown"** with a clearer neutral label:
  - `No safety status yet`
- Keep the neutral state visually neutral, not green/safe.
- When an authenticated user taps **I'm safe** while they have an active help request:
  - show a confirmation dialog first,
  - explain that marking safe will cancel/update the active help request,
  - only proceed if the user confirms.

## Tasks

- Update the Android Home neutral safety badge text to `No safety status yet`.
- Keep green/safe styling only for the actual safe state.
- Check whether the user has an active help request before marking safe.
- If an active request exists, show a confirmation dialog.
- If confirmed, mark the user safe and cancel/update the active help request using the existing request status flow.
- If cancelled, do not change safety status or help request state.

## Acceptance Criteria

- Home no longer shows **"Status unknown"**.
- Neutral state shows **"No safety status yet"** with neutral styling.
- Users without active help requests can mark safe normally.
- Users with active help requests get a confirmation before marking safe.
- Cancelling the dialog makes no changes.
- Confirming marks the user safe and resolves/cancels the active help request consistently.
- Existing Request Help behavior and sync behavior are not broken.

kckagancan:
## Summary

Improves Android Home safety status UX.

## Changes

- Replaces the neutral `Status unknown` badge with `No safety status yet`.
- Adds a confirmation step when marking safe while an active help request exists.
- Keeps guest and normal authenticated mark-safe flows unchanged where applicable.

## Testing

- Verified neutral Home badge behavior.
- Verified guest `I'm safe` behavior stays on Home with login-required message.
- Verified authenticated mark-safe flow with and without active help requests.

Closes #581

### DIFF
--- a/android/app/src/androidTest/java/com/neph/e2e/FakeNephBackend.kt
+++ b/android/app/src/androidTest/java/com/neph/e2e/FakeNephBackend.kt
@@ -208,6 +208,8 @@ class FakeNephBackend {
             route == "/availability/status" && method == "GET" -> handleAvailabilityStatus(token)
             route == "/availability/my-assignment" && method == "GET" -> handleCurrentAssignment(token)
             route == "/help-requests" && method == "GET" -> handleHelpRequestList(token)
+            route.startsWith("/help-requests/") && route.endsWith("/status") && method == "PATCH" ->
+                handlePatchHelpRequestStatus(token, route, body)
             route == "/help-requests/active" && method == "GET" -> handleActiveHelpRequests(uri)
             route == "/location/tree" && method == "GET" -> handleLocationTree(uri)
             route == "/gathering-areas/nearby" && method == "GET" -> handleNearbyGatheringAreas(uri)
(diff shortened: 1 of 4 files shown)


gulcetahtasiz (changes requested):
Thanks, the requested UX changes are implemented and the direction looks good.

Before approval, I think we should handle one safety-critical edge case: when marking safe with an active help request, cancellation is currently queued locally and the safe status can be synced before the help request cancellation is confirmed on the backend. This can temporarily leave the user marked safe while their help request is still active.

Also, the active request check should handle 401 explicitly instead of falling through to the mark-safe dialog.

Could you update this flow so cancellation is confirmed or clearly shown as pending, and add coverage for that behavior?

gulcetahtasiz (approved):
Thanks for addressing the review points. Looks good to me now, merging

PR #586, opened 2026-05-11
fix: notification page timezone handling

### LINKED ISSUE #405: [Bug] Timezone is GMT+0 only.

berat-sayin:
### Description

This PR fixes the remaining timezone issue reported in [#405](https://github.com/bounswe/bounswe2026group6/issues/405) for notification timestamps on the notifications page.

### Changes

- Normalize notification createdAt and readAt values in the backend response to explicit UTC ISO strings.
- Update the shared web timestamp formatter to treat offset-less date-time strings as UTC before formatting them in the user’s local timezone.
- Add unit coverage for notification timestamp normalization, including offset-less timestamp inputs.

### DIFF
--- a/backend/src/modules/notifications/service.js
+++ b/backend/src/modules/notifications/service.js
@@ -74,6 +74,32 @@ function decodeCursor(cursor) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


kckagancan (approved):
Looks fine. Merging.

PR #604, opened 2026-05-12
fix(android): navigate after edited help request submission

### LINKED ISSUE #590: [BUG] Edited help request does not submit or navigate back correctly
### Bug Description

When editing an existing help request, pressing the "Send Help Request" button after selecting all request details does not visibly submit the changes or navigate away from the details page. The user remains on the same screen. If the user manually leaves the screen and later presses the Home button, the app navigates back to the unfinished request details flow as if the edit process is still active.

### Steps to Reproduce

1. Create a help request.
2. Open the request and select "Edit Request".
3. Change/select all request details.
4. Press the "Send Help Request" button.
5. Observe that the page does not close and does not navigate to the Requests page.
6. Manually leave the screen.
7. Press the Home button.
8. Observe that the app navigates back to the request detail selection screen, as if the previous edit flow was unfinished.

### Expected Behavior

After pressing "Send Help Request" while editing, the updated request should be saved and the user should be navigated to the Requests page. The edit/detail selection flow should be cleared from navigation state.

### Actual Behavior

Pressing "Send Help Request" causes no visible screen change. The user stays on the same page, and later navigation can return the user to the unfinished request detail selection screen.

### Severity

High

kckagancan:
## Summary

Fixes the Android help request edit flow.

When an existing help request was edited and submitted, the app could stay on the request form and later restore the unfinished edit flow when navigating back through Home. This PR makes edited request submission complete visibly and clears the transient request form navigation state.

## Changes

- Handles edited help request submission as an explicit update flow.
- Uses the existing local-first update behavior so the edited request is saved locally and queued for sync.
- Navigates to `My Help Requests` after a successful local edit update.
- Clears the `request_help?draftLocalId=...` edit route from navigation state after submit.
- Avoids restoring the old edit screen when the user later presses Home.
- Keeps new request creation, guest flow, offline sync behavior, and onboarding flow unchanged.

Closes #590

### DIFF
--- a/android/app/src/main/java/com/neph/features/requesthelp/presentation/RequestHelpScreen.kt
+++ b/android/app/src/main/java/com/neph/features/requesthelp/presentation/RequestHelpScreen.kt
@@ -934,8 +934,10 @@ fun RequestHelpScreen(
         loading = true
         scope.launch {
             try {
+                val editLocalId = activeDraftLocalId.takeIf { it.isNotBlank() }
+                val existingDraft = editLocalId?.let { RequestHelpRepository.getLocalHelpRequest(it) }
                 if (isLoggedIn) {
-                    val hasActiveRequest = activeDraftLocalId.isBlank() && RequestHelpRepository.hasActiveHelpRequest(sessionToken)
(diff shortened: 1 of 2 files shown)


mehmetcangurbuz08 (approved):
Looks good. Merging.

PR #611, opened 2026-05-12
fix: prevent state restoration on Home navigation

### LINKED ISSUE #549: [Bug] Home page button malfunction on mobile
### Summary
After using the Android system back button during the help request flow, the app gets stuck on the `My Help Requests` screen. Tapping the bottom navigation `Home` tab does not navigate back to the home page.

### Environment
- Platform: Android
- App: NEPH
- Device: Android phone

### Steps to Reproduce
1. Open the app and start from the Home page.
2. Start creating a help request.
3. Fill in the required request details.
4. Submit the help request.
5. The app navigates to `My Help Requests`.
6. Open/edit the active request.
7. Use the Android system back button during this flow.
8. Try to navigate back to the Home page using the bottom navigation `Home` tab.

### Expected Behavior
The app should navigate back to the Home page when the user taps the `Home` tab.

### Actual Behavior
The app remains stuck on the `My Help Requests` screen. The Home tab appears to be tapped/selected, but the page content does not change, so the user cannot return to the Home page.

### Impact
This blocks normal navigation after the user interacts with the help request flow and uses the Android system back button. The user may need to restart the app to reach the Home page again.

### Notes
This looks like a navigation stack/state mismatch after handling the Android system back event. The bottom navigation state may update, but the active route/content does not switch to Home.

AlperKartkaya:
## Summary

  Fixes an Android navigation bug where users could get stuck on `My Help Requests` after using the Android system back
  button during the help-request edit flow. Tapping the bottom-nav Home tab now reliably returns the user to the Home
  screen.

Resolves #549 
  ## Root Cause

  The top-level drawer/bottom-nav navigation helper treated `Home` the same as other destinations and used `saveState =
  true` / `restoreState = true`.

  When navigating from `My Help Requests` back to `Home`, the popped request stack could be saved under the Home route and
  immediately restored, causing `My Help Requests` to remain on top even after the user tapped Home.

  ## Changes

  - Special-cased Home navigation in `AppNavGraph`.
  - Home now behaves as the root destination:
    - first tries to pop back to an existing Home route;
    - if Home is not in the stack, resets the graph to Home;
    - avoids saving/restoring the request stack when Home is selected.
  - Added an Android E2E regression test covering:
    - active help request exists;
    - user opens Requests;
    - user opens Edit Request;
    - user presses Android system back;
    - user taps Home bottom nav;
    - Home content is displayed.

### DIFF
--- a/android/app/src/androidTest/java/com/neph/e2e/AuthenticatedSessionAndroidE2ETest.kt
+++ b/android/app/src/androidTest/java/com/neph/e2e/AuthenticatedSessionAndroidE2ETest.kt
@@ -14,6 +14,11 @@ import com.neph.MainActivity
 import com.neph.features.auth.data.AuthSessionStore
 import com.neph.features.profile.data.ProfileData
 import com.neph.features.profile.data.ProfileRepository
+import com.neph.features.requesthelp.data.RequestHelpContactSubmission
+import com.neph.features.requesthelp.data.RequestHelpLocationSubmission
+import com.neph.features.requesthelp.data.RequestHelpRepository
+import com.neph.features.requesthelp.data.RequestHelpSubmission
+import kotlinx.coroutines.runBlocking
 import org.junit.Rule
 import org.junit.Test
 import org.junit.rules.RuleChain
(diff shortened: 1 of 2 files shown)


delibas-rojhat (approved):
looks good to me, merge.

